PR #2, opened 2016-03-13
Add three exercises, Makefile, travis Crystal install and test, and docs

mhelmetag:
We'll see how travis handles this.

@kytrinyx , Makefile is weird... but I finally got it to work!


kytrinyx:
Nice! Getting CI working is always a great (and sometimes a very frustrating) step.

kytrinyx:
I thought I had given you commit on this repo, but I think GitHub changed their defaults so when I added you it just gave you read access (which for a public repository is not particularly helpful). I've fixed this now.

PR #10, opened 2016-05-29
Fixing description for leap test 2400 so it matches test

derekgottlieb:
I noticed the test code didn't match the description, so updated it to match.

### DIFF
--- a/exercises/leap/example.cr
+++ b/exercises/leap/example.cr
@@ -1,11 +1,11 @@
 class Year
-  def self.is_leap?(year : Int)
+  def self.is_leap?(year : Int32)
     Year.new(year).is_leap?
   end
 
   :private
 
-  def initialize(year : Int)
+  def initialize(year : Int32)
     @year = year
   end
 
@@ -13,7 +13,7 @@ class Year
     divisable_by?(400) || divisable_by?(4) && !divisable_by?(100)
   end
 
-  def divisable_by?(number : Int)
+  def divisable_by?(number : Int32)
     @year % number == 0
   end
 end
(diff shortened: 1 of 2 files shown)


mhelmetag:
@derekgottlieb, thanks! Good find! I have to look into this travis error and fix that first but after that I'll pull this.

derekgottlieb:
I tried taking a stab at fixing the travis error. Switched the year params in the example from Int to Int32 and then 'make test' would at least pass on my machine. Now travis errored out hard and looks like it's no longer finding the crystal binary?

mhelmetag:
That might be fixed by a rebuild. I had that happen before.

derekgottlieb:
Going to close and reopen in the hopes it'll kick off another Travis build. Travis ran clean when I tried it on my fork.

derekgottlieb:
Kicking Travis.

mhelmetag:
Nice!!! Thanks!

kytrinyx:
I'm getting a failure on travis regarding Leap on the open pull request #11 - if we're getting intermittent failures I think we need to dig a bit deeper into what's going on.

derekgottlieb:
I'm curious how Travis builds their Docker image for crystal if the binary sometimes gets left out of the build.

mhelmetag:
I saw that `apt` had failed to grab Crystal correctly once before. I think all the Travis image does is just add the binary to `apt` and then download it (just prevents us from having to do that ourselves and use a sudo container).I

I can see if it ends up being more stable to use a sudo container and manually install crystal in the Travis setup.

mhelmetag:
I think the Leap failure is legit though. That one had to do with a change in Crystal that required a more specific type for seconds... I think.

kytrinyx:
Ahh, ok, that makes sense. Thanks @mhelmetag.

PR #11, opened 2016-05-29
First stab at adding pangram

derekgottlieb:
First attempt at adding an assignment, so please let me know if I've missed something.

### DIFF
--- a/config.json
+++ b/config.json
@@ -10,7 +10,8 @@
     "rna-transcription",
     "bob",
     "raindrops",
-    "leap"
+    "leap",
+    "pangram"
   ],
   "deprecated": [
   ],
--- a/exercises/pangram/example.cr
+++ b/exercises/pangram/example.cr
@@ -1,0 +1,7 @@
+class Pangram
+  NUMBER_OF_ALPHA_CHARACTERS = 26
+
+  def self.pangram?(input : String)
+    input.downcase.gsub(/[^a-z]/, "").chars.sort.uniq.size == NUMBER_OF_ALPHA_CHARACTERS
+  end
+end
--- a/exercises/pangram/pangram_spec.cr
+++ b/exercises/pangram/pangram_spec.cr
@@ -1,0 +1,51 @@
+require "spec"
+require "./pangram"
+
+describe "Pangram" do
+  describe "#pangram?" do
+    it "should reject an empty string" do
+      str = ""
+      Pangram.pangram?(str).should be_false
+    end
+
+    pending "should accept a pangram with only lower case" do
+      str = "the quick brown fox jumps over the lazy dog"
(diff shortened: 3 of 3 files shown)


kytrinyx:
Sweet! This matches the canonical set of tests in https://github.com/exercism/x-common/blob/master/pangram.json so I don't think we're missing any edge cases or anything.

The build is failing... but it's failing on "Leap" which has nothing to do with this. I'll see if I can get whoever dealt with Leap to fix that.

One question about idioms. In Ruby the idiom is to drop the `is_` prefix from predicate methods, since the trailing `?` makes it readable as a question. Does Crystal have any style guides or references that suggest that one or the other is preferable?

kytrinyx:
I'd like the pull request to only have pangram changes in it. Is the second commit the same one as in #10? If so, would you mind rebasing this PR onto the upstream master?

If rebasing is unfamiliar ground, here's one way to do that:

```
git remote add upstream https://github.com/exercism/xcrystal.git
git fetch upstream master
git checkout add-pangram
git rebase upstream/master
git push -fu origin add-pangram
```

derekgottlieb:
I thought I mucked up the git history while trying to rebase. Was hesitant to do the force, but I'll see if I can get that cleaned up and pushed back up. Thanks!

I only started tinkering with Crystal last night, so I'll have to do some searching to see if they have a style guide yet. I'd only gone with is_pangram? since is_leap? was used in the Leap exercise. Will see what I find.

kytrinyx:
Ah, if you do find that they have the same preferences as Ruby `leap?` and `pangram?` vs `is_leap?` and `is_pangram?`, then we'll want to update the leap exercise as well.

derekgottlieb:
The crystal coding style section of their docs is pretty short at this point, so I didn't see any style preference for predicate methods. I did find the following commit from a recent PR that was merged to the core language that would suggest they also prefer to drop the `is_` prefix too:

[Various cleanups #2668](https://github.com/crystal-lang/crystal/pull/2668/commits/f09f1908da8f85a17cec2be812c4f27203b93c85)

Would you prefer I add the  `is_leap?` change in this PR, or open a separate one for that?

kytrinyx:
I'd like to do that in a separate PR.

Thanks for looking into that!

kytrinyx:
I think this looks good, but will leave the final say to @mhelmetag.

Would you mind squashing the commits and making the commit message something like "Implement pangram exercise" or something similar?

derekgottlieb:
Can do!

mhelmetag:
Works for me!

@kytrinyx, thanks for the best practices! I don't think Crystal has a clear styling guide out but since it's so Ruby ish I would warrant a guess that it's not very different.

@derekgottlieb, thanks again for jumping in and helping 😁

Issue #12, opened 2016-05-30
Drop `is_` prefix in predicate methods?


derekgottlieb:
In PR #11, the question was raised whether Crystal had a preference for including prefixes in predicate methods (e.g., `is_pangram?`) or drop them like in Ruby. The coding style section of the Crystal docs is pretty short at this point, so I didn't see any style preference for predicate methods. I did find the following commit from a recent PR that was merged to the core language that would suggest they also prefer to drop the `is_` prefix too:

[Various cleanups #2668](https://github.com/crystal-lang/crystal/pull/2668/commits/f09f1908da8f85a17cec2be812c4f27203b93c85)

Would you like me to go through the existing exercises and drop any prefixes on predicate methods (e.g., `is_leap`)?

mhelmetag:
I'd agree! If it's something that's a best practice in Ruby, it most likely is also in Crystal. By all means change the predicte methods, if you have time!

PR #13, opened 2016-05-30
Dropping prefixes on predicate method is_leap? and fix spelling

derekgottlieb:
Per discussion in #12. I'll put in a separate PR for `Bob` to drop the `is_` prefixes as well.

### DIFF
--- a/exercises/leap/example.cr
+++ b/exercises/leap/example.cr
@@ -1,6 +1,6 @@
 class Year
-  def self.is_leap?(year : Int32)
-    Year.new(year).is_leap?
+  def self.leap?(year : Int32)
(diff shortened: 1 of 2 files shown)


mhelmetag:
I would just make all the changes in one PR that relate to the predicte methods. Seems fine to me... Up to you.

Looks good so far. Thanks for catching that spelling error!

derekgottlieb:
Can do. Would you prefer a commit per example, or just one squashed commit?

mhelmetag:
Ha, I just saw your other PR (I'm traveling and on my phone) so no worries. Thanks for all this!

PR #14, opened 2016-05-30
Dropping prefixes on predicate methods in bob example solution

derekgottlieb:
Per discussion in #12. Didn't notice any more in the other exercises.


mhelmetag:
Thanks!

PR #18, opened 2016-07-19
added gigasecond exercise

elorest:
I added the gigasecond exercise from the ruby track. Am I supposed to add a README somewhere?

### DIFF
--- a/config.json
+++ b/config.json
@@ -7,6 +7,7 @@
   "problems": [
     "hello-world",
     "hamming",
+    "gigasecond",
     "rna-transcription",
     "bob",
     "raindrops",
--- a/exercises/gigasecond/example.cr
+++ b/exercises/gigasecond/example.cr
@@ -1,0 +1,7 @@
+module Gigasecond
+  extend self
+
+  def from(start : Time) : Time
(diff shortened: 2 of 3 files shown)


mhelmetag:
Awesome! Thanks a bunch!

elorest:
I plan on going through and converting a bunch more in the next couple weeks. Is there anything else I need to do other than what I did last night?

mhelmetag:
Not really. That was perfect.

One thing that would be nice is to verify your tests cases against the ones in the exercise json files found [here](https://github.com/exercism/x-common). Those are usually the gold standard that all tracks' test files are based off of.

Also README's for each exercise are found there as well. Those get downloaded when a user fetches an exercise in any language.

I plan on adding a least one exercise this week as well. I'll pick binary search or binary search tree (so we don't overlap) since I've always wanted to do those in crystal anyways.

kytrinyx:
> I plan on going through and converting a bunch more in the next couple weeks.

Sweet! We have this page to help out: http://exercism.io/languages/crystal/contribute (if the page is confusing, would you mind giving a shout? I'd love to make it easier to convert exercises).

PR #19, opened 2016-07-20
atbash_cipher added

elorest:
When will these actually show up on the exercism site?

### DIFF
--- a/exercises/atbash_cipher/example.cr
+++ b/exercises/atbash_cipher/example.cr
@@ -1,0 +1,13 @@
+module Atbash
+  MAPPING = (alp = ('a'..'z').to_a).zip(alp.reverse).to_h
+  extend self
+  def encode(message : String) : String
+    message = message.downcase.gsub(/[^a-z0-9]/, "")
+    encoded_chars = message.chars.map{|c| MAPPING[c]? || c}
+    encoded_string = ""
+    encoded_chars.each_with_index do |c, i|
+      encoded_string += ((i%5 == 0 ? " " : "") + c)
+    end
+    encoded_string.strip
+  end
+end
--- a/exercises/gigasecond/gigasecond_spec.cr
+++ b/exercises/gigasecond/gigasecond_spec.cr
@@ -1,0 +1,35 @@
+require "spec"
+require "./gigasecond"
+
+describe "Gigasecond" do
+  describe "#from" do
+    it "finds gigsecond from 2011_04_25" do
+      result = Gigasecond.from(Time.new(2011, 4, 25, 0, 0, 0))
+      Time.new(2043, 1, 1, 1, 46, 40).should eq result 
+    end
+
+    pending "finds gigsecond from 1977_06_13" do
+      result = Gigasecond.from(Time.new(1977, 6, 13, 0, 0, 0))
+      Time.new(2009, 2, 19, 1, 46, 40).should eq result 
+    end
+
+    pending "finds gigsecond from 1959_07_19" do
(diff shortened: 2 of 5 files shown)


exercises/gigasecond/gigasecond_spec.cr
```diff
@@ -0,0 +1,35 @@
...
+      result = Gigasecond.from(Time.new(2015, 1, 24, 22, 0, 0))
+      Time.new(2046, 10, 2, 23, 46, 40).should eq result 
+    end
+
+    pending "finds gigsecond with full_time with day rollover" do
+      result = Gigasecond.from(Time.new(2015, 1, 24, 23, 59, 59))
+      Time.new(2046, 10, 3, 1, 46, 39).should eq result 
+    end
+
+    pending "test_with_your_birthday" do
```
tejasbubane: missed this?

tejasbubane:
@elorest Your `gigasecond` exercise is already merged with https://github.com/exercism/xcrystal/pull/18. Can you remove the commit for `gigasecond` and send only `atbash cipher`?

exercises/gigasecond/gigasecond_spec.cr
mhelmetag: This is okay as is. It's actually a spot to fill in your own birthday as a test.

But... I do agree that gigasecond already exists...

elorest:
Yes. I'm a little confused why that's in the PR. As it should no longer show up as a difference between our branches. I'll try again.

tejasbubane:
@elorest Might be because you are using `master` branch to send PRs. It is better to use separate branch for each PR you send. A quick google gave me this blogpost http://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/. I can help you on [gitter](https://gitter.im/exercism/dev) in case you find it difficult to get it right.

exercises/atbash_cipher/example.cr
```diff
@@ -0,0 +1,13 @@
...
+  extend self
+  def encode(message : String) : String
+    message = message.downcase.gsub(/[^a-z0-9]/, "")
+    encoded_chars = message.chars.map{|c| MAPPING[c]? || c}
+    encoded_string = ""
+    encoded_chars.each_with_index do |c, i|
+      encoded_string += ((i%5 == 0 ? " " : "") + c)
+    end
+    encoded_string.strip
+  end
```
mhelmetag: It might be worth splitting up the individual functions into their own methods and maybe putting this into a class. Up to you... but I think it would be cleaner.

Like:
```crystal
def encoded_chars(message)
  message.chars.map{|c| MAPPING[c]? || c}
end
```

Or something.

Just looking at the Ruby implementation...

elorest: I agree, although the Ruby implement almost makes the fairly simple problem
look complex by creating functions the do little more than call built in
Ruby functions. I'll go through and add a couple where it makes sense.

elorest:
@tejasbubane I guess I have to add an upstream then do a rebase on the fork. Usually I just rebase from a branch on the same origin. Multiple branches would also have solved this issue although I still am going to want to keep my fork up to date with yours. Thanks for the help.

kytrinyx:
> I guess I have to add an upstream then do a rebase on the fork.

Yeah, that's what I usually do. I try to never work on my fork's master, and then on master I regularly run `git fetch upstream origin && git reset --hard upstream/origin` or something like that, so that every time I start a new branch, it's on the most recent version of the upstream master.

PR #21, opened 2016-07-21
atbash exercise added

elorest:


mhelmetag:
Awesome! Also in response to your question about when this will show up in exercism... this isn't an active track yet. We need 10 exercises. Almost there!

elorest:
@mhelmetag My question was more specifically about why these new ones should show up instead of just the 7? With the 3 I've provided we should be at 10 now though :)

tejasbubane:
Ah, you mean the [languages page](http://exercism.io/languages). If I am not wrong this is because `xapi` is not yet updated http://x.exercism.io/v3/tracks/crystal/todo. Which is because the submodule in https://github.com/exercism/x-api is not yet updated.

config.json
```diff
@@ -12,7 +12,8 @@
     "bob",
     "raindrops",
     "leap",
-    "pangram"
+    "pangram",
+    "atbash_cipher"
```
kytrinyx: This needs to be `atbash-cipher` (hyphen, not underscore)

exercises/atbash_cipher/atbash_cipher_spec.cr
```diff
@@ -0,0 +1,40 @@
+require "spec"
```
kytrinyx: Here too—the directory needs to match the slug, `atbash-cipher`.

kytrinyx:
Sorry—yes, this is my fault. I've had a bit of a hectic week. Deploying now.

PR #22, opened 2016-07-22
largest series product

elorest:


config.json
```diff
@@ -12,7 +12,9 @@
     "bob",
     "raindrops",
     "leap",
-    "pangram"
+    "pangram",
+    "largest-series-product"
+
```
mhelmetag: Probably could delete this blank space.

exercises/largest-series-product/largest_series_product.cr
```diff
@@ -0,0 +1,27 @@
+class Series
```
mhelmetag: I think you didn't mean to add this. The example should be all that you need.

exercises/largest-series-product/largest_series_product.cr
```diff
@@ -0,0 +1,27 @@
...
+    return 1i64 if span == 0 
+
+    product_from_sequence(create_sequences(span))
+  end
+
+  private def create_sequences(span : Int32) : Array(Array(Int64))
+    seq = @series.map_with_index do |n, i|
+      @series[i, span]
+    end
+    seq.select{|s|s.size == span}
```
mhelmetag: I usually like to space out my single line blocks like `{ |s| s.size == span }`. Since Crystal is so much like Ruby, I would follow similar styling practices like [bbatsov's](https://github.com/bbatsov/ruby-style-guide). Just a thought.

elorest:
Resolved.

exercises/largest-series-product/largest_series_product.cr
elorest: Hmmm. I've always thought that looked a bit strange. `{|s| s.size == span}` looks better to me because it provides a clear advantage over `do |s| s.size == span end`. I don't really see why it should be different than something else using {} like a hash. Probably just stuck in my ways after 10 years in ruby. I'll look into the logic behind that a bit more. Thanks for linking that to me.

mhelmetag: Hah I mean... Ruby is generally loose when it comes to styling. I've seen a lot of different looks.

I just did a fair amount of Python and love my PEP8.

It's weird that my RubyMine will indent cases like the first example but I do prefer your second example.

elorest: My vim does it how "right" lol.

asterite: @mhelmetag @elorest I would suggest running `crystal tool format` on the whole project to automatically make the code follow the recommended style.

( Side note: this exercise can also be implemented in a single method chain: https://play.crystal-lang.org/#/r/14n2 )

mhelmetag: @asterite, thanks for the tip! I'll do that. Nice to see a Crystal creator on here 😄 thanks for stopping by

PR #28, opened 2016-07-24
Ran crystal tool format exercises

mhelmetag:
Did some formatting maintenance.


tejasbubane:
Is there a way we can tell travis to check this for each PR? Just like rubocop for ruby?

mhelmetag:
@tejasbubane, definitely. I'll work on this. Shouldn't be that hard to implement.

PR #31, opened 2016-07-26
Correct failure messages in gigasecond exercise

MDaubs:
The subjects in `gigasecond_spec.cr` were reversed which produces failure messages that have the expected value reversed with the "got" value.

Before:
```
  1) Gigasecond #from finds gigsecond from 2011_04_25
     Failure/Error: Time.new(2043, 1, 1, 1, 46, 40).should eq result

       expected: nil
            got: 2043-01-01 01:46:40
```

After:
```
  1) Gigasecond #from finds gigsecond from 2011_04_25
     Failure/Error: result.should eq Time.new(2043, 1, 1, 1, 46, 40)

       expected: 2043-01-01 01:46:40
            got: nil
```


mhelmetag:
Well that's embarrassing... haha. @MDaubs , nice catch.

PR #39, opened 2016-08-07
largest-series-product: Expect exception if input invalid

### LINKED ISSUE #38: largest-series-product: Should it raise on invalid inputs?
Currently, the [largest-series-product tests](https://github.com/exercism/xcrystal/blob/master/exercises/largest-series-product/largest_series_product_spec.cr) expect a value of -1 if the input is invalid in some way (non-digit in digit sequence, negative window size, too-large window size). Should it raise an exception instead? Not sure what the idiomatic way is in Crystal to test for errors, whether it is normal to expect impossible values, but it seems if exceptions are available they can be used.

petertseng:
The current tests expect -1 for invalid inputs, which seems odd; why
force callers to test for a sentinel value, when -1 can never be a
product of digits?

It is understood that x-common's JSON file has expectations of -1 as a
stand-in for the invalid values. However, it is also noted that the same
JSON file specifies that the expectation of -1 should be translated into
an appropriate expectation of failure in the target language.

In Crystal, it seems to be idiomatic to raise an exception for invalid
inputs. Thus, all tests that formerly expected -1 are changed to
`expect_raises(ArgumentError)`.

Note that there is precedent for expecting errors in Hamming where the
inputs are of unequal lengths: there too, the Crystal implementation of
this exercises expects that `ArgumentError` is raised.

The example implementation is updated to match this expectation.

Closes #38

### DIFF
--- a/exercises/largest-series-product/example.cr
+++ b/exercises/largest-series-product/example.cr
@@ -1,14 +1,13 @@
 class Series
+  @series : Array(Int64)
+
   def initialize(series : String)
-    if series =~ /\D/
-      @series = [] of Int64
-    else
-      @series = series.chars.map(&.to_i).map(&.to_i64)
-    end
+    raise ArgumentError.new("Series contains non-digit characters") if series =~ /\D/
+    @series = series.chars.map(&.to_i).map(&.to_i64)
   end
 
   def largest_product(span : Int32) : Int64
-    return -1i64 if span > @series.size || (@series.empty? && span > 0) || span < 0
+    raise ArgumentError.new("Invalid span") if span > @series.size || span < 0
     return 1i64 if span == 0
 
     product_from_sequence(create_sequences(span))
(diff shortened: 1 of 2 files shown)


petertseng:
Hold on, going to amend the commit message; I accidentally wrote "Expect" as "Except"

exercises/largest-series-product/example.cr
```diff
@@ -1,14 +1,13 @@
 class Series
+  @series : Array(Int64)
```
petertseng: It's my understand that this is necesary now that the line reading `@series = [] of Int64` is removed. Removing it would cause something like https://travis-ci.org/exercism/xcrystal/builds/150355827.

mhelmetag: It would seem so... I can't put it in the initializer and I can't seem to omit it. I'm not a Crystal master so for now I think this is a good fix. Might come up with something better but... until then!

Thanks!

PR #44, opened 2016-09-11
Add react exercise

petertseng:
I'm quite partial to this exercise, you see. I've implemented it for Crystal to see what I can do with the language, everything seems to have gone smoothly.

### DIFF
--- a/exercises/react/example.cr
+++ b/exercises/react/example.cr
@@ -1,77 +1,69 @@
-alias CellValue = Int32
+module React
+  alias CellValue = Int32
 
-abstract class Cell
-  @dependents = [] of ComputeCell
+  abstract class Cell
+    @dependents = [] of ComputeCell
 
-  def initialize(initial : CellValue)
-    @value = initial
+    def initialize(initial : CellValue)
+      @value = initial
+    end
+
+    getter :value
+
+    protected getter :dependents
   end
 
-  getter :value
-
-  protected getter :dependents
-end
-
-private class InputCell < Cell
-  def value=(new_value : CellValue)
-    @value = new_value
-    @dependents.each(&.propagate)
-    @dependents.each(&.fire_callbacks)
-  end
-end
-
-private class ComputeCell < Cell
-  alias CallbackID = Int32
-
-  @callbacks_given = 0
-  @callbacks = {} of CallbackID => CellValue ->
-  @last_callback_value : CellValue
-
-  def initialize(@new_value : -> CellValue, *dependencies : Cell)
-    super(@new_value.call)
-    @last_callback_value = @value
-    dependencies.each { |i| i.dependents << self }
+  class InputCell < Cell
+    def value=(new_value : CellValue)
+      @value = new_value
+      @dependents.each(&.propagate)
+      @dependents.each(&.fire_callbacks)
+    end
   end
 
-  def add_callback(&block : CellValue ->) : CallbackID
-    @callbacks_given += 1
-    @callbacks[@callbacks_given] = block
-    @callbacks_given
-  end
+  class ComputeCell < Cell
+    alias CallbackID = Int32
 
-  def remove_callback(id : CallbackID)
-    @callbacks.delete(id)
-  end
+    @callbacks_given = 0
+    @callbacks = {} of CallbackID => CellValue ->
+    @last_callback_value : CellValue
 
-  protected def propagate
-    new_value = @new_value.call
-    return if new_value == @value
-    @value = new_value
-    @dependents.each(&.propagate)
-  end
+    def initialize(input : Cell, &compute : CellValue -> CellValue)
+      initialize(->{ compute.call(input.value) }, input)
+    end
 
-  protected def fire_callbacks
-    return if @value == @last_callback_value
-    @last_callback_value = @value
-    @callbacks.each_value { |c| c.call(@value) }
-    @dependents.each(&.fire_callbacks)
-  end
-end
+    def initialize(input1 : Cell, input2 : Cell, &compute : (CellValue, CellValue) -> CellValue)
+      initialize(->{ compute.call(input1.value, input2.value) }, input1, input2)
+    end
 
-class Reactor
-  # Note that the Reactor does not keep any extra state *in this example solution*.
-  # In other solutions, it's possible that Reactors may keep track of their cells,
-  # take reponsibility for value propagation and callback firing, etc.
+    private def initialize(@new_value : -> CellValue, *dependencies : Cell)
+      super(@new_value.call)
+      @last_callback_value = @value
+      dependencies.each { |i| i.dependents << self }
+    end
 
-  def create_input(initial : CellValue) : InputCell
-    InputCell.new(initial)
-  end
+    def add_callback(&block : CellValue ->) : CallbackID
+      @callbacks_given += 1
+      @callbacks[@callbacks_given] = block
+      @callbacks_given
+    end
 
-  def create_compute(input : Cell, &block : CellValue -> CellValue) : ComputeCell
-    ComputeCell.new(->{ block.call(input.value) }, input)
-  end
+    def remove_callback(id : CallbackID)
+      @callbacks.delete(id)
+    end
 
-  def create_compute(input1 : Cell, input2 : Cell, &block : (CellValue, CellValue) -> CellValue) : ComputeCell
-    ComputeCell.new(->{ block.call(input1.value, input2.value) }, input1, input2)
+    protected def propagate
+      new_value = @new_value.call
+      return if new_value == @value
+      @value = new_value
+      @dependents.each(&.propagate)
+    end
+
+    protected def fire_callbacks
+      return if @value == @last_callback_value
+      @last_callback_value = @value
+      @callbacks.each_value { |c| c.call(@value) }
+      @dependents.each(&.fire_callbacks)
+    end
   end
 end
--- a/exercises/react/example.cr
+++ b/exercises/react/example.cr
@@ -1,0 +1,77 @@
+alias CellValue = Int32
+
+abstract class Cell
+  @dependents = [] of ComputeCell
+
+  def initialize(initial : CellValue)
+    @value = initial
+  end
+
+  getter :value
+
+  protected getter :dependents
+end
+
+private class InputCell < Cell
+  def value=(new_value : CellValue)
+    @value = new_value
+    @dependents.each(&.propagate)
+    @dependents.each(&.fire_callbacks)
+  end
+end
+
+private class ComputeCell < Cell
+  alias CallbackID = Int32
+
+  @callbacks_given = 0
+  @callbacks = {} of CallbackID => CellValue ->
+  @last_callback_value : CellValue
+
+  def initialize(@new_value : -> CellValue, *dependencies : Cell)
+    super(@new_value.call)
+    @last_callback_value = @value
+    dependencies.each { |i| i.dependents << self }
+  end
+
+  def add_callback(&block : CellValue ->) : CallbackID
+    @callbacks_given += 1
+    @callbacks[@callbacks_given] = block
+    @callbacks_given
+  end
+
+  def remove_callback(id : CallbackID)
+    @callbacks.delete(id)
+  end
+
+  protected def propagate
+    new_value = @new_value.call
+    return if new_value == @value
+    @value = new_value
+    @dependents.each(&.propagate)
+  end
+
+  protected def fire_callbacks
+    return if @value == @last_callback_value
+    @last_callback_value = @value
+    @callbacks.each_value { |c| c.call(@value) }
+    @dependents.each(&.fire_callbacks)
+  end
+end
+
+class Reactor
+  # Note that the Reactor does not keep any extra state *in this example solution*.
+  # In other solutions, it's possible that Reactors may keep track of their cells,
+  # take reponsibility for value propagation and callback firing, etc.
+
+  def create_input(initial : CellValue) : InputCell
+    InputCell.new(initial)
+  end
+
+  def create_compute(input : Cell, &block : CellValue -> CellValue) : ComputeCell
+    ComputeCell.new(->{ block.call(input.value) }, input)
+  end
+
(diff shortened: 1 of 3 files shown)


exercises/react/example.cr
```diff
@@ -0,0 +1,72 @@
...
+
+  def fire_callbacks
+    return if @value == @last_callback_value
+    @last_callback_value = @value
+    @callbacks.each_value { |c| c.call(@value) }
+    @dependents.each(&.fire_callbacks)
+  end
+end
+
+class Reactor
```
petertseng: Note that this class does nothing *in this example solution*.

If we anticipate that *all submitted solutions* will do the same, then `create_input` and `create_compute` can be class methods such that they would be called as `React.create_input(1)` rather than `Reactor.new.create_input(1)`. However, in experience with this problem some valid implementations track information about each cell in the reactor. So I propose to leave it like this.

petertseng: I'll add a comment to that effect.

exercises/react/example.cr
```diff
@@ -0,0 +1,76 @@
...
+
+  def create_input(initial : Int32) : InputCell
+    InputCell.new(initial)
+  end
+
+  def create_compute(input : Cell, &block : Int32 -> Int32) : ComputeCell
+    ComputeCell.new(->{ block.call(input.value) }, input)
+  end
+
+  def create_compute(input1 : Cell, input2 : Cell, &block : (Int32, Int32) -> Int32) : ComputeCell
```
petertseng: I wished that I could create a general function of the form `create_compute(*inputs : Cell, &block : Something -> Int32) : ComputeCell`.

However, I was not able to do such a thing. I think the general problem is you cannot write a type for "A function that takes as many inputs as there are values in `inputs`.

Now, you could do `def create_compute(*inputs : Cell, &block : Array(Int32) -> Int32) : ComputeCell` (and the below line would be something like `ComputeCell.new(inputs.map(&.value).to_a}, input1, input2)`. However, now note that we have lost the ability to check that the types match. Someone could create a compute cell with 1 input but whose compute function takes 2 inputs, or vice versa.

So I did it this way to maintain that the compiler can check that the arities match up. Obviously the disadvantage is if we want compute cells that take 3 inputs we have to write a third function, etc.

exercises/react/example.cr
```diff
@@ -0,0 +1,76 @@
...
+    @callbacks_given += 1
+    @callbacks[@callbacks_given] = block
+    @callbacks_given
+  end
+
+  def remove_callback(id : CallbackID)
+    @callbacks.delete(id)
+  end
+
+  def propagate
```
petertseng: I would ideally like `propagate` and `fire_callbacks` to be callable only in this file, though I suppose it's not a big deal. But if there's a way, I'd gladly hear it. file-private is only applicable to top-level declarations, I think.

petertseng: I'll just make these three classes `private class`. Problem solved.

petertseng: Actually, funnily enough, that's doesn't solve the problem. That prevents someone from e.g. creating a cell since they can't do `InputCell.new`, but if they have a `ComputeCell`, they can freely call `propagate` on it. OK, so I guess I'm still looking for a solution on this one (but I wouldn't hold up the PR on it if nobody offers one)

petertseng: Never mind, problem solved for real this time, I think. Add `protected def` for these functions on `Cell`, then override it at `ComputeCell`. Works, and non-`Cell` now can't call these functions.

exercises/react/example.cr
```diff
@@ -0,0 +1,80 @@
...
+
+  def create_input(initial : CellValue) : InputCell
+    InputCell.new(initial)
+  end
+
+  def create_compute(input : Cell, &block : CellValue -> CellValue) : ComputeCell
+    ComputeCell.new(->{ block.call(input.value) }, input)
+  end
+
+  def create_compute(input1 : Cell, input2 : Cell, &block : (CellValue, CellValue) -> CellValue) : ComputeCell
```
petertseng: I wished that I could create a general function of the form create_compute(*inputs : Cell, &block : Something -> CellValue) : ComputeCell.

However, I was not able to do such a thing. I think the general problem is you cannot write a type for "A function that takes as many inputs as there are values in inputs". Therefore, there is nothing I can fill in for Something, above.

Now, you could do def create_compute(*inputs : Cell, &block : Array(CellValue) -> CellValue) : ComputeCell (and the below line would be something like ComputeCell.new(inputs.map(&.value).to_a}, input1, input2). However, now note that we have lost the ability to check that the types match. Someone could create a compute cell with 1 input but whose compute function takes 2 inputs, or vice versa.

So I did it this way to maintain that the compiler can check that the arities match up. Obviously the disadvantage is if we want compute cells that take 3 inputs we have to write a third function, etc.

If you think there is a better way to deal with the problem, let me know.

exercises/react/example.cr
```diff
@@ -0,0 +1,80 @@
+alias CellValue = Int32
```
petertseng: The tests only act on ints. However, you could imagine that the types of the cells could be generic if we want to add more to this problem. Not sure if we want to be doing that.

exercises/react/example.cr
petertseng: > That prevents someone from e.g. creating a cell

false!

`puts typeof(input).new(5).value`

exercises/react/example.cr
petertseng: Okay, I changed my mind about it because it's always possible that someone can do `typeof(existing_cell).new(some_args)`, and if they can do that then I want to make sure `initialize` has to be called with sensible args, and I prefer to do that in one place rather than two.

exercises/react/example.cr
```diff
@@ -0,0 +1,71 @@
...
+
+  @callbacks_given = 0
+  @callbacks = {} of CallbackID => CellValue ->
+  @last_callback_value : CellValue
+
+  def initialize(input : Cell, &compute : CellValue -> CellValue)
+    initialize(->{ compute.call(input.value) }, input)
+  end
+
+  def initialize(input1 : Cell, input2 : Cell, &compute : (CellValue, CellValue) -> CellValue)
```
petertseng: https://github.com/exercism/xcrystal/pull/44#discussion_r78288531 applies.

exercises/react/example.cr
```diff
@@ -0,0 +1,71 @@
...
+  @dependents = [] of ComputeCell
+
+  def initialize(initial : CellValue)
+    @value = initial
+  end
+
+  getter :value
+
+  protected getter :dependents
+  protected def propagate; _propagate; end
```
petertseng: whaaaat? Now that I removed `Reactor` I suddenly had to start doing this protected that delegates to a private for some reason. If I don't do this and just have `protected def propagate` in `ComputeCell`, I get `protected method 'propagate' called for ComputeCell` despite the fact that I didn't change anything else. Why do I have to do this now that I removed `Reactor`? What changed?

petertseng: Ok, it work if I change `@dependents = [] of ComputeCell` to `@dependents = [] of Cell` which  makes sense I guess, but I'm still surprised why it worked when `Reactor` was there.

petertseng: it was *probably* because `ComputeCell` was a private class, and now it is not. I will test a few things to confirm.

petertseng: Yeah.

```crystal
class Thing
  def initialize(@friends : Array(Thing) = [] of Thing)
  end

  protected def hi
    puts "hi from Thing"
  end
end

class SomeThing < Thing
  def hi_friends
    @friends.each(&.hi)
  end
end

class GoodThing < Thing
  protected def hi
    puts "hi from Good Thing"
  end
end

# The below line is a compile error! `protected method 'hi' called for GoodThing`
SomeThing.new([Thing.new, Thing.new]).hi_friends
SomeThing.new([Thing.new, GoodThing.new]).hi_friends

# REQUIREMENT: The below line MUST be a compile error.
# GoodThing.new.hi
# So it does not do to simply make `GoodThing#hi` public.
```

Interesting. And the minimal change to make it work is to make both `Thing` and `GoodThing` private. It does not do to make just one of them private.

Well, I am learning a lot today.

petertseng: Logically, it is also possible to move `hi_friends` up to `Thing` since then `hi_friends` receiver will always be a `Thing` and it is legal for `Thing` to call the `GoodThing#hi` (It is not legal for `SomeThing` to call `GoodThing#hi`). 

In the context of `React`, I don't want to move `hi_friends` (which is `value=`) up to `Thing` (`Cell`), but what I currently have works so it's fine.

exercises/react/example.cr
```diff
@@ -0,0 +1,75 @@
...
+  @dependents = [] of Cell
+
+  def initialize(initial : CellValue)
+    @value = initial
+  end
+
+  getter :value
+
+  protected getter :dependents
+  protected def propagate
```
petertseng: https://github.com/exercism/xcrystal/pull/44#discussion_r78289024 explains why.

exercises/react/example.cr
petertseng: The behaviour is the same as in Ruby, so it is to be expected.

If you attempt to run the above in Ruby, you will succeed if you are just doing `SomeThing.new([Thing.new, Thing.new]).hi_friends` But as soon as you try `SomeThing.new([Thing.new, GoodThing.new]).hi_friends` you will get at run-time: ```protected_methods.rb:13:in `each': protected method `hi' called for #<GoodThing:0x000000024d4ea8 @friends=[]> (NoMethodError)```

You have to do the protected -> private delegation in Ruby as well. So the two languages are doing it the same way. This tells me that this is expected. And now we see that Crystal protects us at compile-time from problems that only happen at run-time in Ruby.

petertseng:
If #45 gets merged before this one, let me know so that I add `react` into `config.json` "exercises" key as well.

And vice versa.

exercises/react/example.cr
petertseng: Figured out that in Crystal `protected` methods are available from "instances in the same namespace (class, struct, module, etc.) as the current type"

So I can avoid the protected->private delegation if I simply throw everything into a module. Let me know if you want me to do that.

exercises/react/example.cr
```diff
@@ -0,0 +1,69 @@
+module React
```
petertseng: Now in a module so that `InputCell` can call `ComputeCell#propagate` but outsiders can't. https://github.com/exercism/xcrystal/pull/44#discussion_r78289024 for context

exercises/react/example.cr
```diff
@@ -0,0 +1,69 @@
...
+
+    @callbacks_given = 0
+    @callbacks = {} of CallbackID => CellValue ->
+    @last_callback_value : CellValue
+
+    def initialize(input : Cell, &compute : CellValue -> CellValue)
+      initialize(->{ compute.call(input.value) }, input)
+    end
+
+    def initialize(input1 : Cell, input2 : Cell, &compute : (CellValue, CellValue) -> CellValue)
```
petertseng: https://github.com/exercism/xcrystal/pull/44#discussion_r78288531 applies (I regret having to do a different overload per arity, but it seems unavoidable if we wish to have the compiler check that the arities match up)

exercises/react/example.cr
```diff
@@ -0,0 +1,69 @@
+module React
+  alias CellValue = Int32
```
petertseng: The tests only act on ints. However, you could imagine that the types of the cells could be generic if we want to add more to this problem. Not sure if we want to be doing that.

petertseng:
Now that #45 is merged: Rebased, and `react` is added to "exercises" key as well as "problems".

exercises/react/react_spec.cr
```diff
@@ -0,0 +1,124 @@
...
+    input = React::InputCell.new(4)
+    input.value = 20
+    input.value.should eq 20
+  end
+end
+
+describe React::ComputeCell do
+  it "calculate initial value" do
+    input = React::InputCell.new(1)
+    output = React::ComputeCell.new(input) { |x| x + 1 }
```
petertseng: I was inconsistent with my names, sometimes using `x` and sometimes using `v`. Please let me fix that

petertseng: fixed.

exercises/react/example.cr
```diff
@@ -0,0 +1,69 @@
...
+
+  private class Cell
```
asterite: Did you mean `abstract` instead of `private`?

exercises/react/example.cr
asterite: @petertseng I believe this is a bug, I just created an issue for it: https://github.com/crystal-lang/crystal/issues/3290

Whenever you find something that looks suspicious, don't hesitate to report it. The language isn't 1.0 yet so there are still some bugs and corner cases. Thank you!

asterite: Actually... after some though, I don't think this can be solved, and the behaviour is probably correct because you are effectively invoking protected methods in unrelated types. Ruby also gives an error for this, so... Nevermind :-)

exercises/react/example.cr
petertseng: Upon review: I think `abstract` does better represent what I want.

petertseng: Thanks!

mhelmetag:
@petertseng , wow! Really great conversations! This is awesome! And it resulted in a ticket in Crystal. So cool!

exercises/react/example.cr
```diff
@@ -0,0 +1,69 @@
...
+      new_value = @new_value.call
+      return if new_value == @value
+      @value = new_value
+      @dependents.each(&.propagate)
+    end
+
+    protected def fire_callbacks
+      return if @value == @last_callback_value
+      @last_callback_value = @value
+      @callbacks.each_value { |c| c.call(@value) }
```
petertseng: I think I've now realized this can be `each_value(&.call(@value))` . But I'm not in a huge hurry to change it here - if necessary I'll use it in what I submit on the site.

PR #46, opened 2016-09-18
attribute the crystal icon, just in case, using the attribution from …

catb0t:
…exercism.io i used when i created it

a step in fixing https://github.com/exercism/exercism.io/issues/3112.


mhelmetag:
Thanks!

Issue #47, opened 2016-11-05
Follow the Crystal convention of placing tests in a spec/ directory


daveyarwood:
According to the Crystal [Spec module documentation](https://crystal-lang.org/api/0.19.4/Spec.html):

> By convention, specs live in the `spec` directory of a project. You can compile and run the specs of a project by running:
>
> ```
> crystal spec
> ```
>
> Also, you can compile and run individual spec files by providing their path:
>
> ```
> crystal spec spec/my/test/file_spec.cr
> ```
>
> In addition, you can also run individual specs by optionally providing a line number:
>
> ```
> crystal spec spec/my/test/file_spec.cr:14
> ```

Are we interested in following this convention? One reason I'm interested in this is that the excellent [vim-crystal plugin](https://github.com/rhysd/vim-crystal) has keybindings to run specs, but it only works if your project follows the convention of having your spec files in a `spec/` directory.

If we're agreed, I'd be happy to contribute a PR updating the existing exercises.

mhelmetag:
Hey there!

Yeah... so we're following the second method (compile and run individual spec files) so that this track can follow the structure of the other track repos (so that the cli and other services can have a standardized way to find all the files).

I'm all for conforming with spec but... I'm not sure if there's a way to change the spec target directory (similar to how xelixir does it [here](https://github.com/exercism/xelixir/blob/master/mix.exs#L9)). Unfortunately, I don't think we can configure that in a `shard.yml` (but if there is, by all means do it).

If you can figure out a way to keep the existing repo structure but also use `crystal spec` that would be awesome (keep in mind that currently we also have most of the tests set to `pending` so to actually run them during CI we do some `sed` magic as well).

daveyarwood:
Hi, thanks for taking the time to explain! I'm not sure I fully understand the problem, yet.

It looks like the [Clojure track](https://github.com/exercism/xclojure) is somehow able to have exercises that each have `src` and `test` directories (following Clojure project convention)... this made me curious, so I dug into the exercism cli, api and trackler repos a little bit. Unless I'm missing something, I'm not sure that there is any requirement for the contents of an exercise folder in a track to follow any particular convention.

I see that we have a Makefile in this repo that manages running the spec files for each exercise. That could certainly be adjusted to `cd` into each directory and run `crystal spec`.

Am I missing something?

mhelmetag:
[...]

PR #48, opened 2016-11-05
Update makefile and leap structure to work with crystal spec

mhelmetag:
@daveyarwood, I took a shot at updating the makefile and structure for leap so that it could use `crystal spec` (this should break all other exercises though).

Is this similar to what you had in mind?

Here's what the output is like now:
```
exercism-xcrystal @ Maxs-MacBook-Pro [max] $ make test-assignment ASSIGNMENT=leap
running formatting check for: leap
moving files around
running tests for: leap
......

Finished in 672 microseconds
6 examples, 0 failures, 0 errors, 0 pending
```

Possible part of the solution for #47


daveyarwood:
Looks legit! I'll put together a PR on top of this branch to update the existing exercises to work with this.

PR #52, opened 2016-11-16
Exercise generator

bmulvihill:
### Overview
Based on https://github.com/exercism/xcrystal/issues/20
Creates an basic exercise generator for the Crystal language track.  The generator lives in the `src` directory.  This generator is based on the ruby generator - but is not a complete copy.  (Beware I am a Ruby dev, and have been getting into Crystal lately, so I have not completely learned all the Crystal best practices).  I have created a sample generator for the hello-world exercise as an example.

### Creating a new Generator
- Navigate to `src/generator/exercises` and create a new generator file (i.e. hello_world.cr).
- This file should contain 2 classes  
  - An exercise generator which must inherit from the ExerciseGenerator class
  - A test case class which inherits from ExerciseTestCase.
- The x-common repo must reside at the same level as the developer's crystal directory, and must contain a canonical-data.json file for the given exercise.

### Running the Generator
From within the xcrystal directory: 

`crystal src/generator/generate.cr hello-world`

Or build a binary
```
make build_generator
bin/generate hello-world
```
### Future Work/Issues
This is just an initial implementation and would like to keep this PR fairly focused on just getting a basic generator implemented. 

Currently a lot of the test suites write the tested method is a describe block (i.e. `describe "#hello"`).  I am not quite sure a clean way to implement this using the canonical data since it doesn't define the tested methods.

Let me know if you have any questions/concerns/comments!

### DIFF
--- a/src/generator/exercises/hello_world.cr
+++ b/src/generator/exercises/hello_world.cr
@@ -2,7 +2,6 @@ require "./exercise_generator"
 require "./exercise_test_case"
 
 class HelloWorldGenerator < ExerciseGenerator
-
   def exercise_name
     "hello-world"
   end
@@ -23,9 +22,9 @@ class HelloWorldTestCase < ExerciseTestCase
 
   def workload
     if !name
-      "HelloWorld.hello.should eq(\"Hello, World\")"
+      "HelloWorld.hello.should eq(\"#{expected}\")"
     else
-      "HelloWorld.hello(\"#{name}\").should eq(\"Hello, #{name}\")"
+      "HelloWorld.hello(\"#{name}\").should eq(\"#{expected}\")"
     end
   end
 
--- a/src/generator/exercises/hello_world.cr
+++ b/src/generator/exercises/hello_world.cr
@@ -1,0 +1,35 @@
+require "./exercise_generator"
+require "./exercise_test_case"
+
+class HelloWorldGenerator < ExerciseGenerator
+
+  def exercise_name
+    "hello-world"
+  end
+
+  def test_cases
+    JSON.parse(data)["cases"].map do |test_case|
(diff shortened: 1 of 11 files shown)


mhelmetag:
@bmulvihill , just curious as to why you closed this out? This looked pretty good!

mhelmetag:
@bmulvihill , I'll look at this after work but I'm glad you reopened it! Looks cool! I would love to add pulling the raw json from x-common's github so we don't have to worry about having x-common locally or it being up to date (I have a branch out that does that so I can merge extend this after).

This is awesome!!

bmulvihill:
Thanks for taking a look!  

I agree about the x-common problem, I just went with what we were doing on the Ruby repo for now since it is a simple solution. 

I also agree a unified generator would be nice, but until that day where all the canonical data adheres to a the same signature/interface creating a simple TestCase class to handle the differences should work.

src/generator/exercises/hello_world.cr
```diff
@@ -0,0 +1,34 @@
...
+
+class HelloWorldTestCase < ExerciseTestCase
+  JSON.mapping(
+    description: String,
+    name: String | Nil,
+    expected: String
+  )
+
+  def workload
+    if !name
```
mhelmetag: I usually like to use `unless` of `if !`. Just a nit...

mhelmetag:
So this is awesome. I think if you're happy with it I'll merge it in and get to working on pulling in the json from github and then knocking out some more exercises. Thanks a bunch again!

src/generator/exercises/hello_world.cr
bmulvihill: I agree, but I have a rebuttal in the case when an else statement is used, I think unless can be confusing, but I could change this to `if name` and switch the logic around.

PR #54, opened 2016-11-19
Don't use json mapping for already parsed JSON

bmulvihill:
Removes parsing the test case JSON and then calling `to_json`.  
This PR changes that to just assign the wanted fields to instance vars.

Open to suggestions on a better way to do this!

### DIFF
--- a/src/generator/exercises/hello_world.cr
+++ b/src/generator/exercises/hello_world.cr
@@ -8,17 +8,21 @@ class HelloWorldGenerator < ExerciseGenerator
 
   def test_cases
     JSON.parse(data)["cases"].map do |test_case|
-      HelloWorldTestCase.from_json(test_case.to_json)
+      HelloWorldTestCase.new(test_case)
     end
   end
 end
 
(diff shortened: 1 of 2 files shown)


mhelmetag:
I haven't done that much json parsing in Crystal but I'm going to look into this a fair bit tomorrow (currently working on a Giphy wrapper). Maybe I'll have some thoughts. But all in all, this looks good!

src/generator/exercises/hello_world.cr
```diff
@@ -8,17 +8,21 @@ class HelloWorldGenerator < ExerciseGenerator
...
   end
 end
 
 class HelloWorldTestCase < ExerciseTestCase
-  JSON.mapping(
-    description: String,
-    name: String | Nil,
-    expected: String
-  )
+  private getter description : JSON::Any
```
mhelmetag: The one thing that I'm a bit confused about with are these private getters where you expose one of them (`description`) anyways.

Is it just a personal preference of less public methods plus the preference of private getter methods over accessing instance variables?

PR #57, opened 2016-11-22
Local or remote metadata files

mhelmetag:
@bmulvihill , kinda a work in progress for adding support for remote (github sourced) metadata test files. It does work but I'm going to try some tests (I hate testing http stuff).

Also fixes the screwiness that messes up the test file states when running `make test`. Sorry about that (woops).

Also I moved a bunch of files around... hope that's okay. I think it makes things cleaner... maybe...

Also tried the remote data file stuff by adding flatten-array.

Let me know how wrong I've done things 😆 

P.S. when this get's 👍 I'll write docs up for this

cc @elorest , because you're (hopefully) a co-maintainer now 😄


config.json
```diff
@@ -118,6 +119,13 @@
       "difficulty": 1,
       "topics": [
       ]
+    },
+    {
+      "slug": "flatten-array",
+      "difficulty": 1,
+      "topics": [
+
```
mhelmetag: I'll fix this with other PR comments

generator/spec/remote_data_file_spec.rb
```diff
@@ -0,0 +1 @@
+require "spec"
```
mhelmetag: Tests will be added (at least one 😆 )

bmulvihill:
@mhelmetag I think this looks good so far!  What are your thoughts of using dependency injection for the DataFile?  i.e. we can pass a flag into the generate script and instantiate our Generator with the correct DataFile (this will respond to the same `.data` method).  This would remove all the DataFile logic from the generator itself.  

Something like this:
```crystal
class ExerciseGenerator
  def initialize(generator = LocalDataFile.new(exercise_name)
    @generator = generator
  end
end
```

mhelmetag:
I think dependency injection is a great idea (especially for testing). I'll fiddle around with that idea tonight.

mhelmetag:
@bmulvihill , have you encountered a problem where one of the params of a function can be several different types (in this instance `LocalDataFile`, `RemoteDataFile` and maybe a `MockDataFile` for testing). How would you type annotate that?

A bit harder than it is in Ruby...

I can see something like `data_file : LocalDataFile | RemoteDataFile | MockDataFile` but then I'll have to include those in the generator... and that kinda defeats the purpose.

I was wondering if you could use a common base type and use that as an annotation but I briefly tried that and couldn't get anywhere.

Thoughts?

bmulvihill:
@mhelmetag would something like this work?

```crystal
class BaseType
end

class A < BaseType
end

class B < BaseType
end

class C
  getter dependency : BaseType

  def initialize(dependency = A.new)
    @dependency = dependency
    p dependency
  end
end

C.new
C.new(A.new)
C.new(B.new)
```

Output:
```
#<A:0x10d878fd0>
#<A:0x10d878fc0>
#<B:0x10d878fb0>
```

bmulvihill:
The other option I think would be to use an `alias`
```crystal
class A
end

class B
end

alias BaseType = A | B

class C
  getter dependency : BaseType

  def initialize(dependency = A.new)
    @dependency = dependency
    p dependency
  end
end

C.new
C.new(A.new)
C.new(B.new)
```

Output:
```
#<A:0x109258fd0>
#<A:0x109258fc0>
#<B:0x109258fb0>
```

mhelmetag:
I'm not 100% sure those fit the bill but definitely helpful. I'll take another crack at this tonight. I have an idea. Thanks!

mhelmetag:
Don't tell me what to do probot... 😆 just kidding.

mhelmetag:
I'm kinda confused as to why there's such a big diff so... I'm reopening this in another PR.

PR #58, opened 2016-11-22
Add forth

bmulvihill:
Adds the forth exercise.

This was a tough (and fun!) one to implement (granted I wrote a Lexer), so I gave it a difficulty of 3 (although I'm not sure how that gets used).  I am open to suggestions on the example implementation.  I am not sure if the input should be an array or just a string, I am leaning towards a string since that is more natural for parsing a language.

```ruby (current)
Forth.evaluate(": DUP-TWICE DUP DUP ; 1 DUP-TWICE")
```
vs
```ruby
Forth.evaluate([": DUP-TWICE DUP DUP ;", "1 DUP-TWICE"])
```

I removed the hardcoding of "tests" in the template, I think that is best left up to the generator implementation. 

It was also a PITA to parse the json for this exercise, so that might be a bit messy.

I also had an issue which I noticed from https://github.com/exercism/xcrystal/pull/57 where I needed to fix empty arrays, so there might be some way to reduce that duplication going forward.

### DIFF
--- a/src/generator/exercises/forth.cr
+++ b/src/generator/exercises/forth.cr
@@ -1,0 +1,57 @@
+require "./exercise_generator"
+require "./exercise_test_case"
+
+class ForthGenerator < ExerciseGenerator
+  def exercise_name
+    "forth"
+  end
+
+  def test_cases
+    test_cases = [] of JSON::Any
+    JSON.parse(data).each do |k, v|
+      v.each do |a, b|
+        test_cases.concat(b) if a.as_s == "cases"
+      end
+    end
+    test_cases.map do |test_case|
+      ForthTestCase.new(test_case)
+    end
+  end
+end
+
+class ForthTestCase < ExerciseTestCase
+  private getter input : JSON::Any
+  private getter description : JSON::Any
+  private getter expected : JSON::Any
+
+  def initialize(test_case)
+    @input = fix_empty_array(test_case["input"])
+    @description = test_case["description"]
+    @expected = fix_empty_array(test_case["expected"])
+  end
+
+  def workload
+    if !(expected == nil)
+      "Forth.evaluate(#{input}).should eq(#{expected})"
+    else
+      <<-WL
+      expect_raises do
+            Forth.evaluate(#{input})
+          end
+      WL
+    end
+  end
+
+  def test_name
+    description
+  end
+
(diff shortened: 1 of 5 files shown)


src/generator/exercises/templates/example.tt
```diff
@@ -3,7 +3,7 @@ require "../src/*"
 
 describe <%= "#{describe_name.inspect} do" %>
 <% test_cases.each_with_index do |test_case, index| %>
-  <%= test_case.pending?(index) %> "tests <%= test_case.test_name %>" do
```
mhelmetag: At some point after this change we should probably refresh the other spec files for consistency (the make task in my PR for looping through all the generators should make refreshes nicer).

Also off topic but related, have you noticed that specs get generated with an extra newline on the first line inside the test block?

mhelmetag: Ahhh derp... that's right. I completely forgot about running format before submitting.

mhelmetag:
> so I gave it a difficulty of 3 (although I'm not sure how that gets used)

I think this will be used for sorting or searching or something. Not sure if any of that is implemented yet.

src/generator/exercises/forth.cr
```diff
@@ -0,0 +1,56 @@
...
+    @input = test_case["input"].as_a.join
+    @description = test_case["description"]
+    @expected = fix_empty_array(test_case["expected"])
+  end
+
+  def workload
+    if !(expected == nil)
+      "Forth.evaluate(#{input.inspect}).should eq(#{expected})"
+    else
+      <<-WL
```
mhelmetag: Just a question... do you prefer heredoc over string literals? If so, why?

bmulvihill: Just preference, and a habit I carried over from Ruby where you can do some neat things with heredocs, not sure if that carries over to Crystal

asterite:
Is this exercise available in other tracks? I couldn't find it.

Or... is there a way to list all exercises regardless of a language?

petertseng:
> Is this exercise available in other tracks? I couldn't find it.

Yup, we can find the answer at: http://x.exercism.io/problems/forth

> is there a way to list all exercises regardless of a language?

The answer to this one is all directories in https://github.com/exercism/x-common/tree/master/exercises

PR #59, opened 2016-11-23
Remove "tests" from template

bmulvihill:
Removes hardcoding of the word 'tests' from the test template and regenerates hello-world and acronym

### DIFF
--- a/exercises/acronym/spec/acronym_spec.cr
+++ b/exercises/acronym/spec/acronym_spec.cr
@@ -2,31 +2,31 @@ require "spec"
 require "../src/*"
 
 describe "Acronym" do
-  it "tests basic" do
+  it "does basic" do
     Acronym.abbreviate("Portable Network Graphics").should eq("PNG")
   end
 
-  it "tests lowercase words" do
+  pending "does lowercase words" do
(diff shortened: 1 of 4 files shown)


exercises/acronym/spec/acronym_spec.cr
```diff
@@ -2,31 +2,31 @@ require "spec"
 require "../src/*"
 
 describe "Acronym" do
-  it "tests basic" do
+  it "does basic" do
     Acronym.abbreviate("Portable Network Graphics").should eq("PNG")
   end
 
-  it "tests lowercase words" do
+  pending "does lowercase words" do
```
mhelmetag: Ah good catch here too to update the pending stuff. I'm a bit annoyed that my current make task mutates the files and doesn't change them back. Oh well. That's fixed in my PR.

src/generator/exercises/acronym.cr
```diff
@@ -29,6 +29,6 @@ class AcronymTestCase < ExerciseTestCase
   end
 
   def test_name
-    description
+    "does #{description}"
```
mhelmetag: Cool and then we can add in whatever verbiage to make the test case descriptions sensical.

mhelmetag:
LGTM 💯

PR #60, opened 2016-11-23
Fix extra space in template

bmulvihill:
Removes extra space from generated specs

### DIFF
--- a/src/generator/exercises/exercise_generator.cr
+++ b/src/generator/exercises/exercise_generator.cr
@@ -9,6 +9,7 @@ abstract class ExerciseGenerator
   end
 
   def generate
+    raise "There are no test cases for #{exercise_name}" if test_cases.empty?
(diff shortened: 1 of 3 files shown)


mhelmetag:
Curious. So if their no test cases the end will be on the same line as the start? At least that's what the test is saying right now...

```shell
expected: "require \"spec\"\nrequire \"../src/*\"\n\ndescribe \"Dummy\" do\nend\n"
          got: "require \"spec\"\nrequire \"../src/*\"\n\ndescribe \"Dummy\" doend\n"
```

bmulvihill:
Hmm good point, it shouldn't matter because we will always have test cases but Ill take a look again

bmulvihill:
I removed the to_s test because I don't feel it added any value, and now an error will be raised if there are no test_cases

mhelmetag:
Works for me! :+1:

PR #61, opened 2016-11-25
Add binary

bmulvihill:
Adds binary exercise


bmulvihill:
@mhelmetag i removed the returns and just output the description, makes more sense I think

mhelmetag:
:100: 2 legit 2 quit

PR #62, opened 2016-11-27
Remove explicit nil checks

bmulvihill:
Use more idiomatic ? syntax

### DIFF
--- a/src/generator/exercises/binary.cr
+++ b/src/generator/exercises/binary.cr
@@ -16,16 +16,16 @@ end
 class BinaryTestCase < ExerciseTestCase
   private getter binary : JSON::Any
   private getter description : JSON::Any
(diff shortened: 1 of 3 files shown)


bmulvihill:
@mhelmetag would you be able to take a quick look at this when you get a moment?

mhelmetag:
@bmulvihill , sorry about that!

Yeah this looks fine. I know that the hash lookup with `?` is idiomatic but is it for type inference as well? Seems like it would be `JSON::Any | Nil` or something like that. Unless the `?` is just syntactic sugar for that same statement.

No complaints here though. Just curiosity.

bmulvihill:
@mhelmetag From my understanding Type? is syntactic sugar for Type | Nil

PR #65, opened 2016-11-30
Add difference of squares

bmulvihill:
Adds difference-of-squares exercise + generator

### DIFF
--- a/config.json
+++ b/config.json
@@ -66,6 +66,12 @@
       ]
     },
     {
+      "slug": "difference-of-squares",
+      "difficulty": 1,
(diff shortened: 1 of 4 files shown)


mhelmetag:
If you're just wrapping a few functions in a namespace, what're your thoughts on `class` vs `module`?

Especially if the class isn't being instantiated?

Like what about:
```crystal
module DifferenceOfSquares
  extend self

  def a; end
  def b; end
  def c; end  
end
```

bmulvihill:
Good idea!  I always forget about `extend self`

PR #67, opened 2016-12-05
Add run-length-encoding

bmulvihill:
Adds run-length encoding exercise + generator

### DIFF
--- a/exercises/run-length-encoding/src/example.cr
+++ b/exercises/run-length-encoding/src/example.cr
@@ -1,0 +1,40 @@
+module RunLengthEncoding
+  extend self
+
+  def encode(str)
+    reader = Char::Reader.new(str)
+    String.build do |io|
+      while reader.has_next?
+        count = 1
+
+        while reader.current_char == reader.peek_next_char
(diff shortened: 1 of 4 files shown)


exercises/run-length-encoding/src/example.cr
```diff
@@ -0,0 +1,40 @@
...
+
+        io << count if count > 1
+        io << reader.current_char
+        reader.next_char
+      end
+    end
+  end
+
+  def decode(str)
+    reader = Char::Reader.new(str)
```
mhelmetag: I'm wondering if this section could be DRY'd out but... it might be more trouble than it's worth...
```crystal
reader = Char::Reader.new(str)
String.build do |io|
  while reader.has_next?
  end
end
```

mhelmetag:
Looks gooood 💯

exercises/run-length-encoding/src/example.cr
bmulvihill: Yeah I originally started down that path but I felt the code became less clear

PR #69, opened 2016-12-12
Add blank solution file to diff-of-squares

bmulvihill:
Noticed this exercise was missing a blank solution file when pulling it from exercism cli


mhelmetag:
Whoops!

PR #72, opened 2016-12-20
Add binary-search-tree

bmulvihill:
Adds binary-search-tree exercise

It currently tests the insert, search, traverse(each), and delete methods.  

Most implementations I have seen in exercism only test insert and traverse, but I think if you are going to implement a Binary Search Tree I feel you should implement the common operations.

I also added tests for additional Crystal features that could be leveraged i.e. making the BST an Enumerable and Iterable (which makes it far more flexible, and allows the programmer to dive a little deeper into the language).  I put comments above them, but maybe I could make it more obvious that they are not *required* tests. What do you think @mhelmetag?

### DIFF
--- a/config.json
+++ b/config.json
@@ -120,8 +120,7 @@
         "data structures",
         "algorithms",
         "trees",
-        "recursion",
-        "iterators"
+        "recursion"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -114,6 +114,17 @@
       ]
     },
     {
+      "slug": "binary-search-tree",
+      "difficulty": 1,
+      "topics": [
+        "data structures",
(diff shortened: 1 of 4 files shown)


bmulvihill:
@mhelmetag Cool, I will move the language specific tests into their own describe block to make it more obvious they are not directly related to implementing a BST.

As far as the JSON file is concerned - I think we could easily revisit that after this goes in.

mhelmetag:
> Cool, I will move the language specific tests into their own describe block to make it more obvious they are not directly related to implementing a BST.

Awesome! That'll work.

> As far as the JSON file is concerned - I think we could easily revisit that after this goes in.

Yep! 👍

PR #79, opened 2017-06-10
Normalize format of config file

kytrinyx:
This will let us script changes to it with less noise in the diffs.


bmulvihill:
Merged https://github.com/exercism/xcrystal/pull/78, which includes the whitespace normalization

PR #91, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

Note that we're aiming for 10 lines and a max-width of 40 columns.
This solution has a max-width of 41, so we may want to adjust it.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,5 @@
+class HelloWorld
+  def self.hello(name : String = "World")
+    "Hello, #{name}!"
+  end
+end
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,5 +1,5 @@
 class HelloWorld
-  def self.hello(name : String = "World")
+  def self.hello(name = "World")
     "Hello, #{name}!"
   end
 end


kytrinyx:
I think we could just get rid of the default and hard code the full 'Hello World' string. @mhelmetag - would you have 5 minutes to fix this? (Also how do you not have commit on this repo? Do you want it?)

mhelmetag:
@kytrinyx , I'm fine with either one! I guess that the hardcoded one is the more canonical "Hello, world" example. And I do have commit access on this repo... I have the ability to merge your PRs? Should be fine then.

bmulvihill:
If it is because the line is too long - do we need to specify the type as a String?  It should work using Crystal type inference.

Example: https://play.crystal-lang.org/#/r/2nje

bmulvihill:
@mhelmetag @kytrinyx I removed the Type, I think this should work.

kytrinyx:
> I do have commit access on this repo

Oh, in that case we're fine :)

Thanks for the tweak @bmulvihill.

I'll leave this to the two of you to merge when you're happy.

PR #98, opened 2018-01-13
Fix generators for exercises with raised exceptions

### LINKED ISSUE #97: New spec failures because of possibly changed method signature/interface
In a [recent automatic build](https://travis-ci.org/exercism/crystal/builds/324853419?utm_source=github_status&utm_medium=notification), I saw a spec failure.

Here's the output from travis:
```
running formatting check for: binary
moving files around
running tests for: binary
Error in line 1: while requiring "./spec/binary_spec.cr"
in spec/binary_spec.cr:42: wrong number of arguments for macro 'expect_raises' (given 0, expected 1, 2, 3, 4)
    expect_raises do
    ^~~~~~~~~~~~~
make[2]: *** [test-exercise] Error 1
make[1]: *** [test-exercises] Error 1
make: *** [test] Error 2
The command "make test" exited with 2.
Done. Your build exited with 1.
```

Seems like we might need to do some fiddling here:
https://github.com/exercism/crystal/blob/3ae242da0e9884518a244387790b8d7270753d61/generator/src/generators/binary.cr#L32

mhelmetag:
Fixes #97

### DIFF
--- a/exercises/acronym/spec/acronym_spec.cr
+++ b/exercises/acronym/spec/acronym_spec.cr
@@ -10,10 +10,6 @@ describe "Acronym" do
     Acronym.abbreviate("Ruby on Rails").should eq("ROR")
   end
 
-  pending "does camelcase" do
-    Acronym.abbreviate("HyperText Markup Language").should eq("HTML")
-  end
-
   pending "does punctuation" do
     Acronym.abbreviate("First In, First Out").should eq("FIFO")
   end
--- a/exercises/binary/spec/binary_spec.cr
+++ b/exercises/binary/spec/binary_spec.cr
@@ -39,37 +39,37 @@ describe "Binary" do
   end
 
   pending "2 is not a valid binary digit" do
-    expect_raises do
+    expect_raises(ArgumentError) do
       Binary.to_decimal("2")
     end
   end
 
   pending "a number containing a non-binary digit is invalid" do
-    expect_raises do
+    expect_raises(ArgumentError) do
       Binary.to_decimal("01201")
     end
   end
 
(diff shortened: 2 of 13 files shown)


bmulvihill:
We would need to regenerate? I'm not sure if we are up to date with all the changes being made to the canonical_data files unfortunately (I haven't checked though).

mhelmetag:
Yeah... We would. That's the issue here. I should just fix that and make a PR for just this... but I started fixing all the other generators and running a full regen and then got too carried away last night... I'll back down and just do this one for now haha.

bmulvihill:
Yeach, if it’s too much we can just fix the specs for now and revisit the generators.   The specs will prevent users from doing the exercises so that’s probably the most important. They still seem to be making a lot of changes 🤷‍♂️

mhelmetag:
Finally!!! 💚 ✔️

bmulvihill:
Hmm I'm not sure I like the int 64 literals, but I see thats the way they are converted from JSON::Any.  

I think that would be confusing maybe for a newcomer to Crystal, and doesn't really add anything to the exercise, its just because we are converting from JSON.  Maybe using JSON::Mapping will allow us to specify the types?  I'll try and take a deeper look tomorrow

mhelmetag:
Ah yeah... that's a good thought. I can look into that as well for these "fixed-up" generators.

bmulvihill:
A simple quick fix we could just gsub out the `_i64`, that way we can get the tests working with the latest version of Crystal

mhelmetag:
That seems like a reasonable quick fix to me.

mhelmetag:
Ah! Thanks for that!

PR #104, opened 2018-04-02
Mark difference_of_squares tests as pending

matthew-healy:
In the majority of exercism crystal exercises all but the first test are marked as pending.
That was not the case for the exercise difference_of_squares.
This commit amends that exercise so it matches the rest.


bmulvihill:
@mhelmetag Not sure what happened, I did try to re-run the generator but all the canonical data has changed over the last few months 😬

PR #105, opened 2018-05-02
trigger prerequisites from related make targets

martinwalsh:
While attempting to figure out how to add an exercise, I noticed that some of the Makefile's build targets were not automatic (e.g. the generator build -> `generator/bin/generator`), so I added a few prerequisites and their corresponding targets, along with a helper target or two.

None of this is strictly necessary, of course. Please feel free to close this PR with impunity if you see fit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,5 @@ bin/generator
 cache/
 generator/lib
 generator/.shards
+generator/bin
 generator/bin/generator
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,6 @@ ci: bin/configlet
 	$(MAKE) test
 
 clean:
-	rm -rf bin/configlet $(GENERATORDIR)/{.shards,bin,cache,lib}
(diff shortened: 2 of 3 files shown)


martinwalsh:
Hi @mhelmetag!

I found the issue with configlet ... my checkout of this repo isn't named `crystal`, since that conflicted with a prior checkout of the source of crystal itself. And it seems that `configlet` [uses the basedir of the project](https://github.com/exercism/configlet/blob/master/track/track.go#L28), by default, to set the track-id which is later used in the api call to check for duplicate uuids. 

I've added configlet's `--track-id` flag explicitly in the `make ci` target, which fixes the issue for me. In theory, I guess we could also update `.travis.yml` to use only `make ci` (or `make clean ci`), but I'm not feeling a strong urge to suggest that at the moment. Perhaps I can make a separate PR with that change and some other suggestions later on, if you're agreeable. 

Thanks again!

Issue #100, opened 2018-01-16
Update generators


mhelmetag:
A lot of problem specifications have been updated to a new format and the generators need to follow suite. Some generators have been updated but I would think that 2/3 of them still need to be checked/updated.

To do this, I would run `make generate-exercises` and see which ones error out and go from there.

You can find the problem specifications for each exercise here: https://github.com/exercism/problem-specifications.

mhelmetag:
I believe all of these generators have been updated.

PR #107, opened 2018-05-12
Refresh flatten_array_spec

chastell:
This refreshes the `flatten_array` spec as per #100.

Cheers from ROSS conf! :tada:

### DIFF
--- a/exercises/flatten-array/spec/flatten_array_spec.cr
+++ b/exercises/flatten-array/spec/flatten_array_spec.cr
@@ -3,23 +3,23 @@ require "../src/*"
 
 describe "FlattenArray" do
   it "no nesting" do
-    FlattenArray.flatten([0, 1, 2]).should eq([0, 1, 2])
+    FlattenArray.flatten([0_i64, 1_i64, 2_i64]).should eq([0_i64, 1_i64, 2_i64])
   end
 
   pending "flattens array with just integers present" do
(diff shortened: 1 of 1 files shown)


chastell:
I mean, this was auto-regenerated, yes, but I should’ve fixed the generator to parse the numbers into `Int32`s rather than make it generate `Int64` literals.

mhelmetag:
Oh yeah! Good catch. Thanks for that.

mhelmetag:
Give me a ping whenever you think you're ready and I'll go ahead and merge this. I think parsing to `Int32` is the way to go.

martinwalsh:
Oh, hi! Didn't notice this PR until now ... looked into this a bit and if I've understood correctly the underlying cause is that [`Array#to_s` calls `inspect` for each of its elements](https://github.com/crystal-lang/crystal/blob/master/src/array.cr#L1661), which would seem to explain why string interpolation results in `Int64` literals.

But, I put together a proposal in #112 which provides an exercise generator that uses `JSON.mapping` internally and bypasses the issue by parsing as `Int32`, more or less, as recommended here. The flatten-array exercise is one of the [example implementations in #112](https://github.com/exercism/crystal/commit/16dd07baa347697e857cab70411bc06c4e505d9f). 

Disclaimer: I'm a crystal n00b and hobbyist who would love to contribute something useful back to the wonderful crystal and exercism communities, so feedback on the PR is most welcome!

chastell:
Oh that’s super cool, I actually tentatively thought about migrating the generators to `JSON.mapping`! Feel free to close this PR in lieu of #112 once that gets accepted. :sparkles:

PR #115, opened 2018-06-11
gigasecond: use Time.utc instead of Time.new in test suite

ilmanzo:
as pointed out here:

https://github.com/crystal-lang/crystal/issues/6173#issuecomment-396190159

in the latest crystal release, Time.new gives local time, so time calculations can be influenced by change of timezone (Daylight Saving Time). Better use Time.utc to avoid similar issues

### DIFF
--- a/exercises/gigasecond/spec/gigasecond_spec.cr
+++ b/exercises/gigasecond/spec/gigasecond_spec.cr
@@ -4,28 +4,28 @@ require "../src/*"
 describe "Gigasecond" do
   describe "#from" do
     it "finds gigsecond from 2011_04_25" do
-      result = Gigasecond.from(Time.new(2011, 4, 25, 0, 0, 0))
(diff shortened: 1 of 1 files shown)


straight-shoota:
It would probably be a good idea to check all exercisms against Crystal 0.25.0.

ilmanzo:
I've ran the entire test suite for checking all 24 exercises against 

```
$ crystal --version
Using compiled compiler at `.build/crystal'
Crystal 0.25.0 [7fb783f7a] (2018-06-11)

LLVM: 3.8.1
Default target: x86_64-pc-linux-gnu

```

and the only exercises impacted by the change is 'gigasecond'.

If, as you suggested, we switch to Time.utc in the spec, it's safe also for users of previous crystal releases.

straight-shoota:
> it's safe also for users of previous crystal releases.
Indeed.

PR #134, opened 2019-02-04
Delete v1 track icon

kytrinyx:
The new Exercism uses custom icons. The v1 icons are no longer in use, and can be deleted.

_If everything is fine, please merge this rather than approving and waiting for me to do so._

Ref: https://github.com/exercism/exercism/issues/4468


bmulvihill:
Looks like we need to update some since CI is now using Crystal `v0.27.1`.  But this should be fine regardless

PR #176, opened 2019-08-18
space-age: improve exercise implementation

dscottboggs:
I believe this implementation more accurately reflects the statically-typed nature of the language, and hopefully teaches the learner some of Crystal's macro primitives.

### DIFF
--- a/exercises/space-age/src/example.cr
+++ b/exercises/space-age/src/example.cr
@@ -1,18 +1,43 @@
-module SpaceAge
-  extend self
+module Space
+  struct Age
+    SECONDS_PER_EARTH_YEAR = 31557600
 
-  @@times = {
-    "Earth"   => 31557600,
-    "Mercury" => 7600543.81,
-    "Venus"   => 19411026.17,
-    "Mars"    => 59359776.78,
-    "Jupiter" => 374222565.14,
-    "Saturn"  => 928792569.65,
-    "Uranus"  => 2652994591.73,
-    "Neptune" => 5196280668.35,
-  } of String => Float64
+    def age_on_mercury
+      @years / 0.2408467
+    end
 
-  def age(planet : String, seconds : Int32) : Float64
-    (seconds / @@times[planet]).round(2)
+    def age_on_venus
+      @years / 0.61519726
+    end
+
+    def age_on_mars
+      @years / 1.8808158
+    end
+
+    def age_on_jupiter
+      @years / 11.862615
+    end
+
+    def age_on_saturn
+      @years / 29.447498
+    end
+
+    def age_on_uranus
+      @years / 84.016846
+    end
+
+    def age_on_neptune
+      @years / 164.79132
+    end
+
+    def age_on_earth
+      @years
+    end
+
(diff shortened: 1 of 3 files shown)


mhelmetag:
Hey there, thanks for the PR!

We're trying not to edit the tests directly. Basically they're generated based with canonical test data (https://github.com/exercism/problem-specifications) and a generator (https://github.com/exercism/crystal/blob/master/generator/src/generators/space_age.cr).

Also, using macros is a great way to solve this (vs a hash)... however, these "examples" are basically just ways to test the test and generator. If people want examples, they should be looking at the ones in exercism.

Does that make sense?

I'd be more than happy to merge a change that includes a change to the test generator but, as is, I believe this goes against what we've been doing (having generators to automagically update the specs).

dscottboggs:
Thanks for the advice, the generator system looks very nice, much easier to make revisions that way than manually. I'll revise my work. :slightly_smiling_face:

dscottboggs:
I've refactored my changes to use the generator system. Let me know what you think.

generator/src/generators/space_age.cr
```diff
@@ -2,14 +2,15 @@ require "../exercise_generator"
 require "../exercise_test_case"
 
 class SpaceAgeGenerator < ExerciseGenerator
+  record TestCases, cases : Array(SpaceAgeTestCase) do
```
mhelmetag: This is pretty cool! Definitely cleaner.

exercises/space-age/src/example.cr
```diff
@@ -1,18 +1,36 @@
-module SpaceAge
-  extend self
+module Space
+  struct Age
+    SECONDS_PER_EARTH_YEAR = 31557600
+    def age_on_mercury
```
mhelmetag: Weren't these using a macro to build out the methods before? I thought that sparked the example change?

generator/src/generators/space_age.cr
```diff
@@ -29,7 +30,7 @@ class SpaceAgeTestCase < ExerciseTestCase
   )
 
   def workload
-    "SpaceAge.#{property}(\"#{input.planet}\", #{input.seconds}).should eq(#{expected})"
+    "Space::Age.from_seconds(#{input.seconds}).#{property}_on_#{input.planet.downcase}.should be_close(#{expected}, 0.01)"
```
mhelmetag: I do like this more OOP approach. I think that's a good example to have.

exercises/space-age/src/example.cr
dscottboggs: Yes, but when you said

> Also, using macros is a great way to solve this (vs a hash)... however, these "examples" are basically just ways to test the test and generator. If people want examples, they should be looking at the ones in exercism.

I thought you meant to remove the macros from the example. I'd be glad to put them back.

generator/src/generators/space_age.cr
dscottboggs: Thanks :slightly_smiling_face:

mhelmetag:
Thanks! Great example of macros, more object oriented-ish programming, and cleaning up our spec generators' JSON handling.

PR #177, opened 2019-08-22
Enable running all specs for each exercise with a compile time flag.

dscottboggs:
This allows a user to run all tests for an exercise by specifying the `RUN_ALL_TESTS` compiler flag, like

```
crystal spec -D RUN_ALL_TESTS
```

Let me know what you think.

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -24,7 +24,13 @@ $ crystal spec
 
 In each test suite all but the first test have been skipped.
 
-Once you get a test passing, you can unskip the next one by changing `pending` to `it`.
+Once you get a test passing, you can unskip the next one by changing
+`{{pending.id}}` to `it` or by defining the RUN_ALL_TESTS compiler flag with:
+
+```bash
+$ crystal spec -D RUN_ALL_TESTS
+```
+
 {{ with .Spec.Credits }}
 ## Source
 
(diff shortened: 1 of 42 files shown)


mhelmetag:
Eh, this goes a bit against the exercism mantra of "simple green" (solving the test one case at a time).

I think it also adds to the complexity of the tests. I'd like them to be as simple as possible.

dscottboggs:
TBH I expected some criticism of this... it just seems like we should be able to use a compile time flag to toggle the `pending` tests, like with `-- --ignored` on the Rust track or `-D CMAKE_CXX_FLAGS="-D EXERCISM_RUN_ALL_TESTS` in the C++ track.

I thought about bringing this up as a PR or issue on Crystal itself, but it seems to go against what `pending` is meant to represent. That made me also think of creating a new macro that wraps `pending` or `it` depending, but I couldn't think of what to call it. Something like

```crystal
macro to_be_done(msg, &block)
  {% if flag? :RUN_ALL_TESTS %}
  it {{msg}} {{block.id}}
  {% end %}
end
```

dscottboggs:
I'm going to close this...if someone else thinks its a good idea, perhaps reopen it, but if it's undesirable, that's ok with me too.

mhelmetag:
If there was some way to make this simpler so that the specs look more normal (`it`, `pending`, `etc`), I'd be game.

PR #179, opened 2019-09-28
triangle: expect new to raise ArgumentError if invalid

petertseng:
Under the principle of failing fast, the author of this commit would
like to argue that it is better design to have the constructor report
when a triangle is invalid, rather than create a Triangle object that is
not actually a triangle, and have the client of the triangle receive
nasty surprises down the road.

This was attempted to be submitted to problem-specifications as
https://github.com/exercism/problem-specifications/pull/1590
however the author realised that it was not possible to justify making
this an error in the canonical-data.json since it directs each track to
make its own decisions about invalid triangles.

Thus it seems it has to be a track-specific decision. Since the Crystal
track is taking the "construct a triangle" route (`Triangle.new`), there
is the possibility of `new` failing.

### DIFF
--- a/exercises/triangle/src/example.cr
+++ b/exercises/triangle/src/example.cr
@@ -3,7 +3,7 @@ class Triangle
 
   def initialize(sides : Tuple(Int32, Int32, Int32) | Tuple(Float64, Float64, Float64))
     @sides = sides.to_a
-    @sides = [] of Int32 if illegal?
+    raise ArgumentError.new("illegal triangle") if illegal?
   end
 
   def equilateral?
--- a/exercises/triangle/spec/triangle_spec.cr
+++ b/exercises/triangle/spec/triangle_spec.cr
@@ -15,7 +15,9 @@ describe "Triangle" do
   end
 
   pending "all zero sides is not a triangle" do
-    Triangle.new({0, 0, 0}).equilateral?.should eq(false)
+    expect_raises(ArgumentError) do
+      Triangle.new({0, 0, 0})
+    end
   end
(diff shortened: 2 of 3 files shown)


exercises/triangle/src/example.cr
```diff
@@ -1,7 +1,7 @@
 class Triangle
-  def initialize(sides)
+  def initialize(sides : Array(Int32))
```
petertseng: note that this is now necessary now that `@sides = [] of Int32` is removed, since otherwise the compiler does not know the type of the instance variable

mhelmetag:
Thanks for merging with master. Sorry about that. Would you mind re-running the generator so that the new specs are part of this PR? I'd like to get out of the habit of changing specs manually. Otherwise this looks great!

petertseng:
Sure. But please understand that running the generator means that decisions must be made about some cases and these decisions are above my pay grade. Since they're above my pay grade, I am obligated to submit them in a separate PR. I would rebase this PR when decisions have been made.

petertseng:
EBWODP: I gave an incorrect reason for why it must be a different PR. It's not because the decisions are above my pay grade; it's because the operation is completely separate from changing the generator

mhelmetag:
Alright, totally my bad for letting this sit for so long. If you could resolve conflicts, I'll merge this ASAP.

petertseng:
OK

petertseng:
OK, resolved.

mhelmetag:
Awesome! This is great! Thanks for the PR

PR #182, opened 2019-11-01
generators/triangle: use JSON::Any#as_i? to format ints/floats

petertseng:
This used to be more interesting before
https://github.com/exercism/crystal/pull/186.
    
Now it's not really useful, but does still generate the same tests.
    
Before this can be merged, it requires justification for why this way
should be used instead of the existing way.

### DIFF
--- a/generator/src/generators/triangle.cr
+++ b/generator/src/generators/triangle.cr
@@ -14,11 +14,11 @@ end
 class TriangleTestCase < ExerciseTestCase
   class Input
     JSON.mapping(
-      sides: Tuple(Int32, Int32, Int32) | Tuple(Float64, Float64, Float64)
+      sides: Tuple(JSON::Any, JSON::Any, JSON::Any)
     )
 
     def to_s(io)
-      io << "{#{sides.map { |s| s.to_i == s.to_f ? s.to_i : s.to_s }.join(", ")}}"
+      io << "{#{sides.map { |s| s.as_i? || s.as_f }.join(", ")}}"
     end
   end
 


exercises/triangle/spec/triangle_spec.cr
```diff
@@ -2,59 +2,79 @@ require "spec"
...
+  pending "two sides are equal" do
     Triangle.new([4, 4, 3]).scalene?.should eq(false)
   end
 
-  pending "Sides that violate triangle inequality are not scalene, even if they are all different" do
+  pending "may not violate triangle inequality" do
     Triangle.new([7, 3, 2]).scalene?.should eq(false)
   end
+
+  pending "sides may be floats" do
```
petertseng: As discussed in the individual commit messages, the test group descriptions such https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/canonical-data.json#L152 are not represented in this test file.

Therefore, there are three cases all named "sides may be floats"

However, the test descriptions have always been missing this context, so fixing this is not within the scope of this PR (and is outside of my interests). For example, exercise `forth` also is affected.

exercises/triangle/spec/triangle_spec.cr
```diff
@@ -2,59 +2,79 @@ require "spec"
...
   end
 
-  pending "All zero sides are illegal, so the triangle is not equilateral" do
+  pending "all zero sides is not a triangle" do
     Triangle.new([0, 0, 0]).equilateral?.should eq(false)
   end
 
-  pending "true if last two sides are equal" do
+  pending "sides may be floats" do
+    Triangle.new([0.5, 0.5, 0.5]).equilateral?.should eq(true)
```
petertseng: as discussed in individual commit message, note test case addition. Since now there may be `Array(Float64)`, consider the student experience when implementing, whereas before there was only `Array(Int32)`. See that the example had to change. If it is too hard for the student to implement this, consider the alternative of changing the generator to not emit these tests.

generator/src/generators/triangle.cr
```diff
@@ -14,11 +14,11 @@ end
 class TriangleTestCase < ExerciseTestCase
   class Input
     JSON.mapping(
-      sides: Array(Int32 | Float32)
+      sides: Array(JSON::Any)
     )
 
     def to_s(io)
-      io << "[#{sides.map { |s| s.to_s }.join(", ")}]"
+      io << "[#{sides.map { |s| s.as_i? || s.as_f }.join(", ")}]"
```
petertseng: if this change is not made, then https://github.com/exercism/crystal/pull/183

generator/src/generators/triangle.cr
```diff
@@ -14,11 +14,11 @@ end
 class TriangleTestCase < ExerciseTestCase
   class Input
     JSON.mapping(
-      sides: Array(Int32 | Float32)
+      sides: Array(JSON::Any)
```
petertseng: if this change is not made, then https://github.com/exercism/crystal/pull/183

mhelmetag:
I haven't tested if the new `to_s` (from the tuple work) in the triangle generator is equivalent to the work done here but... if it's not, I prefer this version. If you still want this change and resolve conflicts, I'd merge ASAP.

petertseng:
OK, here's the treatment I will give this one:

I don't see a point to this one. It results in no changes to the generated tests, so I don't see a reason to use one over the other, unless there are maintenance-related reasons (one version is easier to maintain). Since I am not a maintainer, I don't know what the maintenance-related reasons are, so I cannot provide a reason. And I cannot recommend merging a pull request unless there is a reason for it. So, I resolve the conflicts and close this PR. If there is a valid reason to merge this PR, I leave it to the maintainers to reopen, and edit the commit message to state the valid reason before merging.

petertseng:
Resolved, and confirmed to generate the same tests. Closed.

PR #185, opened 2019-12-30
Wording: Replace function with method

straight-shoota:
Crystal uses methods.

### DIFF
--- a/exercises/darts/README.md
+++ b/exercises/darts/README.md
@@ -1,6 +1,6 @@
 # Darts
 
-Write a function that returns the earned points in a single toss of a Darts game.
+Write a method that returns the earned points in a single toss of a Darts game.
 
 [Darts](https://en.wikipedia.org/wiki/Darts) is a game where players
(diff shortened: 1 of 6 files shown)


mhelmetag:
I'll look into this! I believe these readmes are copied from a canonical exercise readme so I'm not sure it's right to edit them.

straight-shoota:
I figured as much. But the correct/idiomatic term for Crystal is method. Crystal *can* expose functions using the `fun` keyword, but what you would expect from normal Crystal code are regular methods defined using `def`.

I don't know whether there is any value in keeping the upstream nomenclature, but for Crystal it would be better to use the term "method"

I'm not sure whether there is some value

mhelmetag:
Sure! I completely agree. I just don't really want to maintain basically patches to the canonical readmes.

I should be able to get back to you with a yay or nay or more info later this week.

straight-shoota:
Is this still open for consideration?

mhelmetag:
Because these are from static readmes, I'd like to keep them the way they are (unedited and from our list of problem metadata).

PR #186, opened 2020-01-02
Use tuples for the triangle exercise instead of arrays

Steffan153:
Changed the Triangle exercise to use tuples instead of arrays.

### DIFF
--- a/exercises/triangle/spec/triangle_spec.cr
+++ b/exercises/triangle/spec/triangle_spec.cr
@@ -2,59 +2,79 @@ require "spec"
 require "../src/*"
 
 describe "Triangle" do
-  it "true if all sides are equal" do
-    Triangle.new({2, 2, 2}).equilateral?.should eq(true)
+  it "all sides are equal" do
+    Triangle.new({2.0, 2.0, 2.0}).equilateral?.should eq(true)
   end
 
-  pending "false if any side is unequal" do
-    Triangle.new({2, 3, 2}).equilateral?.should eq(false)
+  pending "any side is unequal" do
+    Triangle.new({2.0, 3.0, 2.0}).equilateral?.should eq(false)
   end
 
-  pending "false if no sides are equal" do
-    Triangle.new({5, 4, 6}).equilateral?.should eq(false)
+  pending "no sides are equal" do
+    Triangle.new({5.0, 4.0, 6.0}).equilateral?.should eq(false)
   end
 
-  pending "All zero sides are illegal, so the triangle is not equilateral" do
-    Triangle.new({0, 0, 0}).equilateral?.should eq(false)
+  pending "all zero sides is not a triangle" do
+    Triangle.new({0.0, 0.0, 0.0}).equilateral?.should eq(false)
   end
 
-  pending "true if last two sides are equal" do
-    Triangle.new({3, 4, 4}).isosceles?.should eq(true)
+  pending "sides may be floats" do
+    Triangle.new({0.5, 0.5, 0.5}).equilateral?.should eq(true)
   end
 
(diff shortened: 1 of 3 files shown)


Steffan153:
In Crystal, normally, a tuple would be fitting for this. You have three sides--and always three--doesn't that sound like a usecase for a tuple? 

BTW, This started when I was talking with a mentor on Exercism, and he brought it up that tuples should be used, and I agreed...

> Also, if no one else has made a pull request yet: the test suite is incorrect, and should be passing `Tuple(Int32, Int32, Int32)` rather than `Array(Int32)`.

tenebrousedge:
The canonical spec is written in a language which does not have first-class support for tuples, so it doesn't seem like a good argument for or against the use of them: it cannot express the concept, so it uses the closest analogue. Per the mathematical definition, triangles have precisely three sides. In Crystal, this is expressed as `Tuple(Int, Int, Int)`, where a generic polygon might be `Array(Int)`. Using the latter introduces the possibility of runtime errors, and various students have included code in their solutions to test for these runtime errors. This is properly a (semantic) typing error in the test requirements: passing more or fewer sides should result in a compiler error, not a runtime error. Having students attempt to work around potential deficiencies in the test code does teach them something, but it would be far better to teach them how to manipulate Tuples, which are a first-class datatype in Crystal that is not currently used in any other problem. Triangle is, in fact, the ideal problem for illustrating the differing uses between `Tuple` and `Array`.

I agree that changing the generator is a better idea.

Steffan153:
@tenebrousedge: I [already changed the generator](https://github.com/exercism/crystal/pull/186/commits/2062c355c0f6d21b264a2a2bb1a461cda80308dc).

mhelmetag:
The last part would be to regen the spec file and commit that (unless there are no changes). That okay?

Steffan153:
Hmm... I did that and now everything is floats?

mhelmetag:
Why did you make the type a union type with `Float32`? Was that causing problems with the generator?

Steffan153:
The [canoncial data](https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/canonical-data.json) has tests with floats...

exercises/triangle/spec/triangle_spec.cr
```diff
@@ -2,59 +2,79 @@ require "spec"
 require "../src/*"
 
 describe "Triangle" do
-  it "true if all sides are equal" do
-    Triangle.new([2, 2, 2]).equilateral?.should eq(true)
+  it "all sides are equal" do
+    Triangle.new({2.0, 2.0, 2.0}).equilateral?.should eq(true)
```
petertseng: Here is an opinion from a random bystander (me). This opinion **is not intended** to be a substitute for maintainer opinion.

I think it would be preferable if these were written as `{2, 2, 2}` instead of `{2.0, 2.0, 2.0}`, and only use floats for the specific tests that use floats. It is the approach I took in https://github.com/exercism/crystal/pull/182. It still has not been reviewed.

Steffan153: Good idea! I found a way to do that, done.

Thing is, previously, the exercise didn't have any floats at all. ?

tenebrousedge: > Thing is, previously, the exercise didn't have any floats at all. ?
That is a correct statement. Perhaps you had another related question?

mhelmetag:
The canonical might've changed between now and the last regen. It looks like you just have to update the example src files to work with both float and int and then you should be home free.

Steffan153:
Done!

mhelmetag:
Works for me! Thanks for this!

PR #197, opened 2020-11-22
fetch-configlet: Accept lowercase location

petertseng:
https://github.com/exercism/configlet/issues/173

See https://github.com/exercism/crystal/pull/194 - all four PRs are required for build to pass again:

* https://github.com/exercism/crystal/pull/195
* https://github.com/exercism/crystal/pull/196
* https://github.com/exercism/crystal/pull/197
* https://github.com/exercism/crystal/pull/198

PR #196, opened 2020-11-22
crystal tool format

petertseng:
A few unnecessary backslashes are removed.

bob
binary
hamming
series

It's been confirmed that running all these generators produces no
changes, so it's safe to remove these backslashes with no further
changes.

```sh
for x in bob binary hamming series; do crystal run ./generator/generator.cr -- $x; done
```

See https://github.com/exercism/crystal/pull/194 - all four PRs are required for build to pass again:

* https://github.com/exercism/crystal/pull/195
* https://github.com/exercism/crystal/pull/196
* https://github.com/exercism/crystal/pull/197
* https://github.com/exercism/crystal/pull/198

PR #195, opened 2020-11-22
meetup: Time.utc

petertseng:
Time.new with those particular args removed in 0.33.0 (2020-02-14)

See https://github.com/exercism/crystal/pull/194 - all four PRs are required for build to pass again:

* https://github.com/exercism/crystal/pull/195
* https://github.com/exercism/crystal/pull/196
* https://github.com/exercism/crystal/pull/197
* https://github.com/exercism/crystal/pull/198

PR #198, opened 2020-11-22
matching-brackets: escape backslashes in input string; set all but first to pending

petertseng:
It's necessary to do this because one input has backslashes in it, and
not escaping them means that they are treated as escape sequences along
with the character immediately following them.

That would cause the generator and `crystal tool format` to disagree on
what the spec should look like. The formatter will remove the backslash
before `\left` and `\mathrm` because `\l` and `\m` are not escape
sequences listed in
https://crystal-lang.org/reference/syntax_and_semantics/literals/string.html

It's admitted that the presence/absence of the backslashes is immaterial
to this problem, since the backslash is not a bracket character that is
being checked for pairedness.

However, failure to escape the backslashes angers those who know LaTeX
because then the string is no longer valid LaTeX.

Therefore, it is better to escape them.

Notice how in the JSON file, it was also necessary to escape the
backslashes.
https://github.com/exercism/problem-specifications/tree/master/exercises/matching-brackets

See https://github.com/exercism/crystal/pull/194 - all four PRs are required for build to pass again:

* https://github.com/exercism/crystal/pull/195
* https://github.com/exercism/crystal/pull/196
* https://github.com/exercism/crystal/pull/197
* https://github.com/exercism/crystal/pull/198

### DIFF
--- a/exercises/practice/matching-brackets/spec/matching_brackets_spec.cr
+++ b/exercises/practice/matching-brackets/spec/matching_brackets_spec.cr
@@ -67,6 +67,6 @@ describe "MatchingBrackets" do
   end
 
   pending "complex latex expression" do
-    MatchingBrackets.valid?("\left(\begin{array}{cc} \frac{1}{3} & x\\ \mathrm{e}^{x} &... x^2 \end{array}\right)").should eq(true)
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):

ErikSchierboom:
@petertseng Would you prefer a rebased merge, preserving the commits, or a squash merge?

petertseng:
I think it is better to preserve these commits.

ErikSchierboom:
Done. Thanks!

PR #261, opened 2022-02-02
Makefile: ignore shared

petertseng:
Otherwise, it will try to test shared/.docs as an exercise, when it
isn't one.

All other items in IGNOREDIRS can go away because none of those
directories are directories under exercises

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-IGNOREDIRS := "^(\.git|.crystal|docs|bin|img|script)$$"
+IGNOREDIRS := "^shared/"
 EXERCISESDIR ?= "exercises"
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Is this safe to merge with the failing tests?

petertseng:
If the desire is not to have failing tests, then merge #195 first, and then I would rebase this on main.

petertseng:
but, the state of the world is made no worse by merging this PR first. Tests already fail on `main` because both this PR and #195 are needed, so it doesn't matter which of the two is merged first.

ErikSchierboom:
I'm happy to merge.

PR #268, opened 2022-04-16
Implement Clock exercise for Crystal

marksiemers:


exercises/practice/clock/.meta/config.json
```diff
@@ -0,0 +1,20 @@
...
+  "contributors": [],
+  "files": {
+    "solution": [
+      "src/clock.cr"
+    ],
+    "test": [
+      "spec/clock_spec.cr"
+    ],
+    "example": [
+      ".meta/src/example.cr"
```
marksiemers, line 15: The `solution` and the `example` are the exact same code - I wasn't sure if this was desired. If it should be something different, please let me know.

exercises/practice/clock/.meta/tests.toml
```diff
@@ -0,0 +1,39 @@
+# This is an auto-generated file. Regular comments will be removed when this
+# file is regenerated. Regenerating will not touch any manually added keys,
+# so comments can be added in a "comment" key.
+
+[1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
+description = "basic"
```
marksiemers, line 6: Looks like this is a left-over copy/paste from acronym. I'll try to clean that up.

Although, I have not been able to get the generator to work. Is this file supposed to be generated locally or does it happen in another part of the workflow?

generator/src/generators/clock.cr
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def test_cases
+    TestCaseGroup(ClockTestCase).from_json(data).cases[0].cases
+  end
+end
+
+class ClockTestCase < ExerciseTestCase
+  class Input
+    JSON.mapping(
```
marksiemers, line 16: Anytime I try to generate something I get this error:
```
 16 | JSON.mapping(
           ^------
Error: undefined method 'mapping' for JSON:Module
```

This code is taken from the `acronym` example - perhaps it is the version of Crystal that I have locally?
```
Crystal 1.3.2 [932f193ae] (2022-01-18)

LLVM: 10.0.0
Default target: aarch64-apple-darwin
```

marksiemers (commented):
This requires a little cleanup, I'm looking for guidance on a few things:
1. Should the "example" and the "solution" be the same code? Or should one be removed?
2. Is the `tests.toml` file supposed to be generated locally and committed?
3. Is there a way to correct the error I'm seeing with `JSON.mapping`?

Thank you

ErikSchierboom:
@exercism/reviewers anyone familiar with Crystal that can take a look at this?

marksiemers:
This will change a bit after https://github.com/exercism/crystal/pull/269 is merged and the generators are working again.

Leaving this PR open for now, as a lot of the changes will remain the same.

exercises/practice/clock/src/clock.cr
```diff
@@ -0,0 +1,71 @@
+class Clock
+  HOURS_PER_DAY    = 24
```
petertseng, line 2: since this file is provided to the students, it would leave the students with no work to do if this file already contained a complete solution. all functionality should be removed from this file, leaving only the signatures of the public functions that will be tested by the test suite. Those should be just the function declaration and parameters, with no body

PR #269, opened 2022-04-18
Get exercise generator working

marksiemers:


ErikSchierboom (commented):
It looks good to me, but I don't know Crystal. Maybe someone from @exercism/reviewers does?

marksiemers:
@ErikSchierboom - it looks like there hasn't been a lot of activity in the track as of late.

I'm seeing a handful of committers over the past 9-ish months:
@ee7 @neenjaw @richard-sistern @petertseng 

Who are the members or maintainers of this track? Are they still interested in maintaining it?

neenjaw:
I wrote the test runner in crystal but I don't purport myself to be fluent in crystal -- crystal just happens to be lovely to work in. I can probably provide as much sanity check as @ErikSchierboom. I would not consider myself a maintainer of this track.

ee7:
I'm not a Crystal maintainer either - sorry. My commits in this repo were part of Exercism-wide work, and others are better placed than me to review this PR.

It looks like Peter has done the most Crystal-specific work lately:
https://github.com/exercism/crystal/graphs/contributors?from=2020-04-20&to=2022-04-20&type=c

But the answer is probably that the track doesn't have particularly active, dedicated maintenance right now.

petertseng:
Now, the organisation members can see via https://github.com/orgs/exercism/teams/crystal that **there are no current maintainers** for the Crystal track (non-members will not be able to view that link, so unfortunately they do not have much recourse but to take my word for it). I don't have any advice as to how to remedy that.

As for this PR, the changes to generator/shard.yml and generator/src/exercise_generator.cr are exactly what I expected to see and I'm not aware of what needed to change in generator/src/templates/example.tt so I have no specific advice to offer for that specific file. My advice is if this PR fixes the generators, might as well make it official by re-enabling the generator tests in https://github.com/exercism/crystal/blob/b87a3fa9201ea083fa8f8930d757b76ed01ffd75/.github/workflows/tests.yml#L22-L23.

ErikSchierboom:
@marksiemers The suggestion made by @petertseng in the comment above (re-enabling the CI test for the generators) seems like a good way to verify that this works. Would you mind applying that suggestion?

marksiemers:
@ErikSchierboom and @petertseng - I made changes that got `make test-generator` working locally. And I've also updated the todo in the config.

So hopefully it will all work in the running workflows too.

.github/workflows/tests.yml
```diff
@@ -19,5 +19,4 @@ jobs:
       - uses: actions/checkout@dcd71f646680f2efd8db4afa5ad64fdcba30e748
       - name: Run tests
         run: make test-exercises
-        # TODO: fix generators
-        #run: make test-generator
+        run: make test-generator
```
ErikSchierboom, line 22: You can't have two `run` keys. To run multiple commands, the `|` character needs to be used. See https://docs.github.com/en/github-ae@latest/actions/using-workflows/workflow-commands-for-github-actions#example-grouping-log-lines

petertseng: that's my fault for making the comment this way, then

petertseng, line 22: In that case, I think we had better just make it a separate Step (in the GitHub Actions sense of the word)

```suggestion
      - name: Run generator tests
        run: make test-generator
```

again, my apologies for making the original comment wrong

ErikSchierboom: @marksiemers Could you apply peter's suggestion? That should fix things

hiljusti (approved):
LGTM

I will claim experience enough in Crystal to approve it

ErikSchierboom:
https://github.com/exercism/crystal/pull/269/checks looks good. Thanks @marksiemers

PR #274, opened 2022-05-07
Update Makefile and README.md

marksiemers:
Some make commands were failing for various reasons.
Fix Makefile to get commands working.

Update README to provide working examples using 'make'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,11 +44,15 @@ $(GENERATORBIN):
 $(GENERATORBIN)/generator: $(G_SRCS) | $(GENERATORBIN)
 	@crystal build $(GENERATORDIR)/generator.$(FILEEXT) -o generator/bin/generator
 
-build-generator: $(GENERATORBIN)/generator
+.PHONY: build-generator
+build-generator:
+	@echo "building binary for generator: $(GENERATOR)"
+	@cd $(GENERATORDIR) && (shards check || shards install) && mkdir -p bin && crystal build generator.cr -o bin/generator
 
-generate-exercise: $(GENERATORBIN)/generator
+.PHONY: generate-exercise
+generate-exercise:
 	@echo "generating spec file for generator: $(GENERATOR)"
-	@generator/bin/generator $(GENERATOR)
+	@cd $(GENERATORDIR) && (shards check || shards install) && crystal generator.cr $(GENERATOR)
 
 generate-exercises:
 	@for generator in $(GENERATORS); do GENERATOR=$$generator $(MAKE) -s generate-exercise || exit 1; done
@@ -57,7 +61,7 @@ test-generator:
 	@echo "running formatting check for generators"
 	@crystal tool format --check $(GENERATORDIR)
 	@echo "running generator tests"
-	@cd $(GENERATORDIR) && shards install && crystal spec
+	@cd $(GENERATORDIR) && (shards check || shards install) && crystal spec
 
 test:
 	@echo "running all the tests"
--- a/README.md
+++ b/README.md
@@ -7,15 +7,18 @@ Exercism problems in Crystal.
 ### All Exercises
 
 Test all exercises with:
-```bash
-$ make test-exercises
+```shell
+make test-exercises
 ```
 
 ### Single Exercises
 
 Test single exercises with:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -44,11 +44,13 @@ $(GENERATORBIN):
...
-build-generator: $(GENERATORBIN)/generator
+build-generator:
+	@echo "building binary for generator: $(GENERATOR)"
+	@cd $(GENERATORDIR) && (shards check || shards install) && crystal build generator.cr -o bin/generator
 
-generate-exercise: $(GENERATORBIN)/generator
+generate-exercise:
 	@echo "generating spec file for generator: $(GENERATOR)"
-	@generator/bin/generator $(GENERATOR)
+	@cd $(GENERATORDIR) && (shards check || shards install) && crystal generator.cr $(GENERATOR)
```
marksiemers, line 53: Is it better to build the generator binary, then call it?

Crystal really provides both options - building first, or executing inline.

ErikSchierboom: What is the downside of executing inline? As generators are only rarely being called, performance should probably not be an argument (not sure if it is).

hiljusti: I'd either:

1. Run it in an interpreted mode
2. Give it a make rule (so you at least get "dirty bit" rebuilds)

But I think running interpreted is fine.

Makefile
```diff
@@ -44,11 +44,13 @@ $(GENERATORBIN):
 $(GENERATORBIN)/generator: $(G_SRCS) | $(GENERATORBIN)
 	@crystal build $(GENERATORDIR)/generator.$(FILEEXT) -o generator/bin/generator
 
-build-generator: $(GENERATORBIN)/generator
+build-generator:
```
hiljusti, line 47:
```suggestion
.PHONY: build-generator
build-generator:
```

On the off-chance that someone might actually put a `build-generator` file or executable here, or rename the target to something that matches a file, this would prevent Make thinking it should only run this target if the file modification time was updated.

Makefile
```diff
@@ -44,11 +44,13 @@ $(GENERATORBIN):
...
+	@echo "building binary for generator: $(GENERATOR)"
+	@cd $(GENERATORDIR) && (shards check || shards install) && mkdir -p bin && crystal build generator.cr -o bin/generator
 
-generate-exercise: $(GENERATORBIN)/generator
+generate-exercise:
```
hiljusti, line 51:
```suggestion
.PHONY: generate-exercise
generate-exercise:
```

(Same reasoning as `build-generator`)

Makefile
marksiemers: @hiljusti - Do you think it is too early to run in interpreted mode?

> The interpreter is currently on an experimental phase, with lots of [missing bits](https://github.com/crystal-lang/crystal/issues/11555). The reason to merge it at this early stage is to enable a proper discussion of interpreter-related PRs and speed up its development a bit. For those willing to try it out, we welcome interpreter-related issues taking into consideration the already known issues aforelinked.

Source: https://crystal-lang.org/2021/12/29/crystal-i.html

hiljusti: Maybe, but idk, I'd wait until seeing a problem before changing it. It seems to be working fine so far, and it is just running a single file so if it's working now it might just keep working forever? (i.e. it's not some crazy huge thing with complex dependencies that would stress the edges of their interpreter)

hiljusti (approved):

marksiemers:
@ErikSchierboom - I think this is ready to merge.

@hiljusti - I think some clarification is needed about running crystal, I would say there are three ways to run it:
1. "Inline" (unsure the real term) - `crystal your_file.cr`
2. "Interpreted - `crystal i your_file.cr` (I think this allows a REPL like experience)
3. "Compiled" - `crystal build --release your_file.cr && ./your_file`

I'm assuming and hoping we're all on board with using "inline" mode - since it is working and not a high risk of causing problems.

ErikSchierboom:
@marksiemers I'll wait for @hiljusti's approval, as I don't really know Crystal

hiljusti:
Let's do it!

Thanks for your patience, getting to these when I can.

PR #306, opened 2023-01-15
[re-architectured generator]: Using ecr instead of liquid

meatball133:


bin/generator.sh
```diff
@@ -0,0 +1,20 @@
...
+# error prints an error to STDERR and exits.
+
+if [ -z "$1" ] ; then
+    echo "usage: ./bin/template.sh exercise-slug "
+    exit 1
+fi
+
+slug="$1"
+
+cat "./exercises/practice/${slug}/.meta/test_template.ecr" > "./bin/templates/template.ecr"
```
ErikSchierboom, line 14: Shouldn't this be the other way around?

```suggestion
cat "./bin/templates/template.ecr" > "./exercises/practice/${slug}/.meta/test_template.ecr"
```

bin/generator_help.cr
```diff
@@ -100,4 +89,37 @@ class GeneratorHelp
...
+
+  def to_snake(input)
+    result = ""
+    input.each_char do |x|
+      result += x.upcase == x ? "_#{x.downcase}" : x
+    end
+    result
+  end
+
+  def to_captlized(input)
```
ErikSchierboom, line 101:
```suggestion
  def to_capitalized(input)
```

ErikSchierboom (approved):

bin/generator.sh
meatball133: No it should be that way, the script takes the template for each exercise and move it to a set folder.

bin/generator.sh
```diff
@@ -0,0 +1,24 @@
...
+if [ -z "$1" ] ; then
+    echo "usage: ./bin/template.sh exercise-slug "
+    exit 1
+fi
+
+slug="$1"
+
+if test ! -d example_directory; then
+  mkdir example_directory
+fi
```
ErikSchierboom, line 16: Did you want to create the `bin/templates` directory here?

PR #308, opened 2023-01-16
[Deprecate react]: Exercise broken in test-runner

meatball133:


ErikSchierboom (commented):
What is the issue? Why is the test runner failing?

meatball133:
I have no clue why the test-runner is falling, but it is something with the json thingy. 
All tests pass localy.
I spent the weekend on improving the test tooling for the test runner, but I got stuck in making a pr to the crystal-lang and the generator.
I am aiming for next week to have that new test-runner test tools to be done.

I could also dig into fixing this but I have quite a lot to do, and it feels like the exercise isnt draging that many students any way.

ErikSchierboom:
> I am aiming for next week to have that new test-runner test tools to be done.

Okay! We can un-deprecate later.

ErikSchierboom (approved):

PR #323, opened 2023-01-20
[Ci]: make ci faster

### LINKED ISSUE #313: Seperate Ci tests
## Background

The ci test is made of 2 parts, one that checks so the "examples" works with the test files, and then another one that compares so the file that the generator would generate is the same test file that is there.
How the current ci is setup is that it first runs the "checks so the "examples" works with the test files" part and then the other part after that.
The ci tests takes around 5 minutes to run at the moment, the test part takes around 4 minutes and the templates part takes 1 minute, but that is with only around 20 exercises having a template. 

Thereby to be able to reduce the amount of time the ci would require to run.
Would it be beneficial if they can run at the same time since they don't rely on each other.

## What?

Separate so the generator test is separate from the exercise tests ci to reduce the amount of time it takes to run.

Here is the current file which handles the tests:

```yml
name: Tests
on:
  pull_request:
  push:
    branches:
      - main
  schedule:
    # Weekly.
    - cron: '0 0 * * 0'

jobs:
  exercises:
    name: Check exercises
    runs-on: ubuntu-latest
    # idea from https://github.com/actions/starter-workflows/blob/master/ci/crystal.yml
    container:
      image: crystallang/crystal
    steps:
      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
      - name: Run tests
        run: make test-exercises
      - name: Run generator tests
        run: ./bin/test_specs
```

## Want more info?

You can write in this issue or message me on slack, my username there is: meatball.

meatball133:
resolves #313

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   exercises:
     name: Check exercises
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     # idea from https://github.com/actions/starter-workflows/blob/master/ci/crystal.yml
     container:
       image: crystallang/crystal
@@ -21,7 +21,7 @@ jobs:
         run: make test-exercises
   generator:
     name: Check Generator Templates
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     container:
       image: crystallang/crystal
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,5 +19,12 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
       - name: Run tests
         run: make test-exercises
+  generator:
+    name: Check Generator Templates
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -19,5 +19,12 @@ jobs:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
       - name: Run tests
         run: make test-exercises
+  generator:
+    name: Check Generator Templates
+    runs-on: ubuntu-latest
```
ErikSchierboom, line 24: I would recommend to pin to a specific version:

```suggestion
    runs-on: ubuntu-22.04
```

This helps prevent potential unforeseen breakages in the future. p.s. I know the other action in this workflow also uses `ubuntu-latest`, so feel free to merge this and update both in a follow-up PR.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

ErikSchierboom (approved):

PR #331, opened 2023-01-25
[New practice exercise]: Word Count

meatball133:


ErikSchierboom (approved):

exercises/practice/word-count/src/word_count.cr
```diff
@@ -0,0 +1,5 @@
+class WordCount
```
ErikSchierboom, line 1: Should this be: 

```suggestion
module WordCount
```

ErikSchierboom (approved):
I don't necessarily need to be a contributor though, but 🤷

PR #348, opened 2023-02-21
[New practice exercise]: Pythagorean Triplet

ryanplusplus:


exercises/practice/pythagorean-triplet/spec/pythagorean_triplet_spec.cr
```diff
@@ -0,0 +1,32 @@
...
+  pending "returns all matching triplets" do
+    PythagoreanTriplet.triplets_with_sum(90).should eq([[9, 40, 41], [15, 36, 39]])
+  end
+
+  pending "several matching triplets" do
+    PythagoreanTriplet.triplets_with_sum(840).should eq([[40, 399, 401], [56, 390, 394], [105, 360, 375], [120, 350, 370], [140, 336, 364], [168, 315, 357], [210, 280, 350], [240, 252, 348]])
+  end
+
+  pending "triplets for large number" do
+    PythagoreanTriplet.triplets_with_sum(30000).should eq([[1200, 14375, 14425], [1875, 14000, 14125], [5000, 12000, 13000], [6000, 11250, 12750], [7500, 10000, 12500]])
```
meatball133, line 30: I recoomend splitting these lines, having super long lines can be hard to read, thereby I instead recommend to write it like this:
```crystal
expected = <data>
PythagoreanTriplet.triplets_with_sum(30000).should eq(expected)
```

I have written a formatter which can format arrays, tuples and hashes. Which will be runned automatically when running the generator but it only works when assigning to a variable.

exercises/practice/pythagorean-triplet/src/pythagorean_triplet.cr
```diff
@@ -0,0 +1,5 @@
+module PythagoreanTriplet
+  def self.triplets_with_sum(sum : Int32) : Int32
```
meatball133, line 2: The output here doesn't seem right, shouldnt it return `Array(Array(Int32))`?

meatball133 (commented):
Nice work! ✨

I left some small comments.

Also I have a bit of a release shedule of practice exercises for the track. I plan to release around 5 exercises per month. The reasson is not to dump a lot of exercises but instead give time to think through thier implimentation.

That doesnt mean I don't allow other people to write practice exercise, it is just that I rather want people to disscus with me before. Also since I may have an idea how I think it should be implimented, for example so does some exercises make sense to impliment in more oop shape comapred to other once. I am generally pretty open to allow people to add some exercises which are on my list. It is just in the future so you are aware. This exercise am I just glad that you put the time in, because I don't like it.

I think the difficlty rating is about right, the example looks good.

Also if you like to contribute with other stuff aswell so do I have a list of stuff which has to be done, in any case I should probably open them up as issues.

exercises/practice/pythagorean-triplet/spec/pythagorean_triplet_spec.cr
ryanplusplus: Sure, makes sense, I'll update the generator.

exercises/practice/pythagorean-triplet/src/pythagorean_triplet.cr
ryanplusplus: Whoops, good catch!

ryanplusplus:
> Nice work! sparkles
> 
> I left some small comments.

Thanks!

> Also I have a bit of a release shedule of practice exercises for the track. I plan to release around 5 exercises per month. The reason is not to dump a lot of exercises but instead give time to think through thier implimentation.
> 
> That doesnt mean I don't allow other people to write practice exercise, it is just that I rather want people to disscus with me before. Also since I may have an idea how I think it should be implimented, for example so does some exercises make sense to impliment in more oop shape comapred to other once. I am generally pretty open to allow people to add some exercises which are on my list. It is just in the future so you are aware. This exercise am I just glad that you put the time in, because I don't like it.
> 
> Also if you like to contribute with other stuff aswell so do I have a list of stuff which has to be done, in any case I should probably open them up as issues.

Sure, that would be great. I checked and didn't see any issues so figured I'd just start with an easy exercise to get started. I'd be happy to help through some track infrastructure or exercises if you can get some information into tickets.

meatball133:
Well here is a list of stuff that needs to be done:

March exercises

- Bottle Song
- Pig Latin
- Sublist
- Hangman
- Bank Account (dependent on if the exercises gets conical data until then)

Feel free to do the first 3 exercises on this list under March, I sorta want to implement Hangman by myself if that is okay.


Help review concepts is something that will be needed for a while.

Writing concept, I sorta want to author the early concepts, but branching of concepts have I no problem that somebody else writes in the early stages, those concepts include: 

- Binary & octal & hexadecimal
- BigInt, Bigfloat, Bigdecimal
- Setters and getters

If you are interested is also a possibility to co-author the early concepts.

There is a need to add more ci scripts, if that is something you like doing is that an option.

I have also thought about replacing `to_capitalized` with the built-in method `camelcase` and `to_snake` with `underscore`.
I am unsure it is a quite good first issue, so I may reserve it to first-timers only.

If you like to work on the test runner so would I love to make the user able to write p and puts it in the online editor and make that show up in the result. If you like to do this one so requires a deep discussion with me and perhaps Erik about how that should be approached.

That issue is also time pressured since the crystal test-runner is running a very outdated version, it requires an update and then 1.8 gets released will speed up testing occur to be able to update the test runner quickly, thereby we can't have other work going on at the same time.

meatball133 (approved):

ryanplusplus:
> Well here is a list of stuff that needs to be done:
> 
> March exercises
> 
>     * Bottle Song
> 
>     * Pig Latin
> 
>     * Sublist
> 
>     * Hangman
> 
>     * Bank Account (dependent on if the exercises gets conical data until then)
> 
> 
> Feel free to do the first 3 exercises on this list under March, I sorta want to implement Hangman by myself if that is okay.
> 
> Help review concepts is something that will be needed for a while.
> 
> Writing concept, I sorta want to author the early concepts, but branching of concepts have I no problem that somebody else writes in the early stages, those concepts include:
> 
>     * Binary & octal & hexadecimal
> 
>     * BigInt, Bigfloat, Bigdecimal
> 
>     * Setters and getters
> 
> 
> If you are interested is also a possibility to co-author the early concepts.
> 
> There is a need to add more ci scripts, if that is something you like doing is that an option.
> 
> I have also thought about replacing `to_capitalized` with the built-in method `camelcase` and `to_snake` with `underscore`. I am unsure, it is a quite good first issue, so I may reserve it to first-timers only.
> 
> If you like to work on the test runner so would I love to make the user able to write p and puts it in the online editor and make that show up in the result. If you like to do this one so will that require a deep discussion with me and perhaps Erik about how that should be approached.
> 
> That issue is also time pressured since the crystal test-runner is running a very outdated version, it requires an update and then 1.8 gets released will speed up testing occur to be able to update the test runner quickly, thereby we can't have other work going on at the same time.

Thanks for the info. I'm definitely not an expert in Crystal and my goal is to learn more by contributing. I'm happy to work on whatever exercises you'd like me to and will leave the rest to you.

I'll have a look at your concept exercises soon and will begin by working on Bottle Song, Pig Latin, and Sublist then go from there as I get more comfortable with Crystal and the track's tooling.

Thanks for your help!

PR #362, opened 2023-03-06
[Ci]: Replace makefile ci with bash ci

meatball133:
The goal of this pr/project is to remove all makefile code. The purpose of such a change is to make the track easier to maintain since it would reduce the number of languages (Crystal, bash, jq, makefle) into 3. This script doesn't really allow for any new functionality the old script couldn't do.

Another goal of this pr is to start to test with multiple versions of crystal, this has been for a long time been my plan. Since it has caused issues in the past that only the latest version is tested, here is a recent example: [link](https://github.com/exercism/crystal/pull/312). Thereby will this new script test 3 versions, 1.2, 1.6, 1.7. When the tooling gets updated to 1.8, so is my plan to instead test 1.6, 1.7, 1.8 and perhaps 1.5. Thereby dropping official support for 1.2 entirely.

The test runner runs 1.2.2 to give some context.


ErikSchierboom (commented):
Excellent idea! You can simplify this a bit more by parameterizing on the installed version.
See https://github.com/exercism/rust/blob/main/.github/workflows/tests.yml#L96-L100 and https://github.com/exercism/rust/blob/main/.github/workflows/tests.yml#L113-L117 for how to do that.

meatball133:
@ErikSchierboom do you intend like this?

ErikSchierboom (approved):
Yes, this is perfect!

PR #368, opened 2023-03-08
[New practice exercise]: Killer sudoku helper

meatball133:


exercises/practice/killer-sudoku-helper/.meta/test_template.ecr
```diff
@@ -0,0 +1,20 @@
...
+    end
+    <%- end -%>
+    <%- else -%>
+    <%= status()%> "<%-= cases["description"] %>" do
+    killer_sudoku_helper = <%-= to_capitalized(cases["property"].to_s) %>.new(<%-= cases["input"]["cage"]["sum"] %>, <%-= cases["input"]["cage"]["size"] %>, <%-= cases["input"]["cage"]["exclude"] %> of Int32)
+    killer_sudoku_helper.solve.should eq <%-= cases["expected"] %>
+    end
+    <%- end %>
+<% end %>
+end
```
ryanplusplus, line 20: Could you indent this to improve readability? Since it's run through the crystal formatter after generation, the extra whitespace shouldn't matter, but it would make it easier for me to follow this.

exercises/practice/killer-sudoku-helper/.meta/src/example.cr
```diff
@@ -0,0 +1,15 @@
+class Combinations
+  getter sum, size, exclude
```
ryanplusplus, line 2:
```suggestion
  private getter sum, size, exclude
```

ryanplusplus (commented):
Looks good, just a couple of nitpicks

ErikSchierboom (approved):
Approving pending the changes suggested by ryan being applied

ryanplusplus:
Looks like the `darts` generator failed. There was a [minor change](https://github.com/exercism/problem-specifications/commits/main/exercises/darts/canonical-data.json) to the problem specification shortly before your last commit, but I don't see why that spelling change would have broken anything.

meatball133:
The generator ci is super sensative to changes on problem spec. If more test cases are added it wont freak out, but as fast as some one change or gets in a different order, it fails.

I just pushed a pr to fix it any way.

exercises/practice/killer-sudoku-helper/.meta/test_template.ecr
meatball133: Some of them are written without being that formatted, but as you said the generator is setup to not care about how it is written. This is very nice since some languages generators are super sensative to this because they use intentation (python).

Although I think we should have some "standard" way of formatting these documents. If you are up to come up with a design so are you free to, otherwise I will do it at some point.

Also I changed my mind of hangman and replaced it with this one. Since hangman doesnt have conanical data.
Bank account will likely also be a no go.

If you want you want to do another exercise this month, you can do one of these once:

Affine Chiper
Bowling
Yatch
Linked list
Tournament
Two buckets
Book store

If you dont want so can I do it.

ryanplusplus: I'll plan to grab one of those.

PR #378, opened 2023-03-18
[Concepts]: Strings

### LINKED ISSUE #377: [Concepts]: Add strings
## Goal

The goal of this exercise is to teach the student about strings in crystal.

## Learning objectives

- Know what a string is
- How to define a string
- How to concat strings 
- How to convert to and from strings to other types
- Interpolation with strings
- How to get the length of a string
- Indexing of a string
- Escaping
- How to make a multi line string

## Out of scope

- Most strings method
- Heradoc
- Precent strings literals

## Concepts

`strings`:

- .size
- interpolation (`#{}`)
- Escaping(`\n`)
- indexing([i])
- concatination of strings
- Conversion between types

## Prerequisites

- Classes

meatball133:
@angelikatyborska
@glennj

resolves #377


concepts-wip/strings/introduction.md
```diff
@@ -0,0 +1,157 @@
+# Strings
+
+Crystal has a type known as [`String`][strings] which is used to represent text.
+In Crystal is a String a sequence of unicode characters.
```
glennj, line 4:
```suggestion
Crystal strings are sequences of unicode characters.
```

concepts-wip/strings/introduction.md
```diff
@@ -0,0 +1,157 @@
...
+hello + " World"
+# => "Hello World"
+```
+
+When wanting to concat a string with a non-string type so can you use the method `to_s` to convert the type to a string.
+
+```crystal
+"Hello " + 1.to_s
+# => "Hello 1"
+```
```
glennj, line 30:
```suggestion
```
You talk about `to_s` just below.

concepts-wip/strings/introduction.md
```diff
@@ -0,0 +1,157 @@
...
+
+```crystal
+"Hello World".size
+# => 11
+```
+
+## Indexing
+
+Indexing is when you want to get a specific character from a string.
+To get a character from a string you can use the `[]` method with the index of the character you want to get.
```
glennj, line 90: A suggestion:

```suggestion
To get a character from a string you can use familiar bracket notation.

~~~~exercism/note
`[]` is actually implemented as a String instance method, where the index is the method argument.
~~~~
```

concepts-wip/strings/about.md
```diff
@@ -0,0 +1,178 @@
...
+
+```crystal
+name = "Hello World"
+name[name.size - 1]
+# => 'd'
+
+name[-1]
+# => 'd'
+```
+
```
glennj, line 112: In the About document, you might want to talk about other ways to index a string: ranges, regex, substring
```crystal
icr:1> s = "Hello World"
 => "Hello World"
icr:2> s[1]
 => 'e'
icr:3> s[1..4]
 => "ello"
icr:4> s[/e.*o/]
 => "ello Wo"
icr:5> s["nope"]
Unhandled exception: Nil assertion failed (NilAssertionError)
```

glennj (changes requested):

concepts-wip/strings/about.md
meatball133: This is something I have thought about in how to deal with this. Since a range is actually a datatype in Crystal and you can do a range over more than just numbers. So personally I thought it would make more sense to introduce that later when chars have been explained. Ranges are a bit complicated in how they should be introduced since it deals with numbers, strings, chars and arrays. 

I mean we could do a short introduction and then do a full introduction in a later concept. What do you think?

Substring, that I think could make sense to introduce here so I think that is a good idea.

Regex, I believe deserves its own concept, and I think perhaps a mention could happen but not a deeper dive with examples. Do you think that makes sense?

glennj: If it goes into About, you don't need to put it in the exercise. It can be a "teaser" for later concepts. Range and Regex are definitely their own concepts. So, yes, I agree.

ErikSchierboom (approved):

PR #383, opened 2023-04-01
[New practice exercise]: Say

ryanplusplus:


meatball133 (approved):
Sorry for taking some time, I had a busy weekend with an important deadline that came before exercism. That is now handed in.

Anyway, it looks good, just one thing and that is to add yourself as author.

ryanplusplus:
> Sorry for taking some time, I had a busy weekend with an important deadline that came before exercism. That is now handed in.
> 
> Anyway, it looks good, just one thing and that is to add yourself as author.

No problem at all, take your time to review. Thanks for catching my mistake with the authors field (again).

PR #388, opened 2023-04-08
[Concepts]: add ternary operator

### LINKED ISSUE #387: [Concepts]: Add Ternary Operator
## Goal

The goal of this exercise is to teach the student about the ternary operator in crystal.

## Learning objectives

- Know about ternary operator

## Out of scope

- Everything that Conditionals has as out of scope

## Concepts

`Ternary-operator`:

- condition ? true_branch : false_branch

## Prerequisites

- Conditionals

meatball133:
@angelikatyborska
@glennj

resolves #387

### DIFF
--- a/concepts-wip/ternary-operator/.meta/config.json
+++ b/concepts-wip/ternary-operator/.meta/config.json
@@ -1,0 +1,5 @@
+{
+  "blurb": "Crystal has a ternary operator, which allows you to write a conditional expression in a single line.",
+  "authors": ["meatball133"],
+  "contributors": []
+}
--- a/concepts-wip/ternary-operator/about.md
+++ b/concepts-wip/ternary-operator/about.md
@@ -1,0 +1,48 @@
+# Introduction
+
+## The ternary operator
+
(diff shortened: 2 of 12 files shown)


glennj:
Does this need to be a separate concept, or can it be mentioned in a Conditionals concept?

meatball133:
I mean, it could be, I did this mostly to avoid the concept of the conditional for becoming too long and I don't know if the exercise suits using ternary. 

I am also planning to add if as a suffix to the about.md on conditionals (will do that this weekend). But I mean if you feel like adding this won't make the concept of the conditional too large, I think we can perhaps merge them

glennj:
IMO it only needs to be very short:

````
## Ternary Operator

Crystal implements the familiar ternary operator: `condition ? expression_if_true : expression_if_false`

```crystal
1 == 0 ? "equal" : "not equal"
# => "not equal"
```
````

meatball133:
Yeah, I think we can merge them.

I will have to look over it later this week.

PR #400, opened 2023-04-13
[Concepts]: Add while loops

### LINKED ISSUE #399: [Concepts]: Add while loops
# Goal

The goal of this exercise is to teach the student about while loops in crystal.

## Learning objectives

- Know about while and until statements
- Know how to create a basic while loop
- Know how to skip to the next iteration of a loop
- Know how to exit a loop early

## Out of scope

- For loops

## Concepts

`While-loops`:

- while
- until
- next
- break

## Prerequisites

- Conditionals

meatball133:
@angelikatyborska
@glennj

resolves #399

### DIFF
--- a/concepts/return/introduction.md
+++ b/concepts/return/introduction.md
@@ -1,0 +1,28 @@
+# Return
+
+In Crystal so does always the last expression in a method return its value.
+Although sometimes you need to return a value before the last expression.
+For this you can use the `return` keyword.
+Code written after a `return` keyword will not be executed.
+
+```crystal
+def speed_limit(road_number)
+    if road_number == 1
+        return 50
+        puts "This will not be executed"
+    end
+    if road_number == 2
+        return 80
+        puts "This will not be executed"
+    end
+    100
+end
+
+p speed_limit(1) # => 50
+p speed_limit(2) # => 80
+p speed_limit(3) # => 100
+```
+
+```exercism/note
+The `return` keyword should be omitted when the last expression in a method is the value that should be returned.
+```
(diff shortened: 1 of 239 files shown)


concepts/while-loops/about.md
```diff
@@ -0,0 +1,147 @@
...
+a = while true
+  break 1
+end
+
+p a
+# => 1
+
+a = while true
+  1
+end
```
marksiemers, line 138: This looks like in infinite loop

concepts/return/introduction.md
```diff
@@ -0,0 +1,28 @@
+# Return
+
+In Crystal so does always the last expression in a method return its value.
```
marksiemers, line 3: The introduction and about look the same, is this intentional?

marksiemers (commented):
This looks pretty good.

I may not understand how a concept is supposed to be added to Exercism, but it looks like a lot of duplicative content (about and introduction).

concepts/while-loops/about.md
meatball133: Yeah, that is not intended to be that way. Thanks

concepts/return/introduction.md
meatball133: The introduction is about how the concept looks before you start the exercise and about is what it looks after. I quite often add some extra tidbit in about but other than that so are they the same

meatball133:
Hi!

Yes, there is some duplicate-looking content but they are slightly different and each file has a different purpose.

glennj (changes requested):
Just a few documentation suggestions.

The chosen exercise does not really demonstrate much of the concept material (no early return, no break/next). I'm going to look around for an exercise that might be more specifically relevant.

ErikSchierboom (approved):

PR #410, opened 2023-04-27
Fix typo in Roman Numerals

Steffan153:
Simple typo I noticed while solving it.

Unrelated to this PR, but I'm interested in contributing to this track. There's not many issues open, but is it open to new practice exercises? Or are there other things, like building analyzers?


ryanplusplus:
Good find, but this file is generated from upstream here:
https://github.com/exercism/problem-specifications/blob/main/exercises/roman-numerals/description.md?plain=1

Would you mind putting a pull request together upstream?

Steffan153:
Oh of course, I'm dumb.

PR #411, opened 2023-04-29
[New practice exercise]: Matrix

ryanplusplus:


meatball133 (approved):
Awesome ! ✨ 

I think the maintainer bit will have to wait until next week, since I think Jeremy won't reply over the weekend.

For exercises for next month, if you want you can pick a bit of more or less whatever you feel is appropriate. With the small expectation that exercise without test files or exercise marked as deprecated on the problem spec, could be good to have a discussion before adding.

If you don't know what you feel is appropriate to add, I can make a list but I also leave the option for yourself to add what you feel is best. My personal plan is to add Minesweeper so all exercise out of that one is just for you to choose.

PR #412, opened 2023-05-06
[New practice exercise]: Two Bucket

ryanplusplus:


config.json
```diff
@@ -677,6 +677,14 @@
         "prerequisites": [],
         "difficulty": 5
       },
+      {
+        "slug": "two-bucket",
+        "name": "Two Bucket",
+        "uuid": "20c471f8-e45e-4313-a23a-85d4a36c7185",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5
```
meatball133, line 686: I think this exercise is quite hard and since it uses structs and enums I think perhaps the difficulty should be put as 6.

meatball133 (commented):
Looks good!
Just one tiny comment

config.json
ryanplusplus: Good thought, fixed!

meatball133 (approved):
Awesome!✨

PR #418, opened 2023-05-28
Small-fixes for navigation-computer

ryanplusplus:
When trying this concept exercise out, I noticed a typo and a small formatting inconsistency.


meatball133 (commented):
Looks good!

If you could apply that spell fix to the exemplar as well so would that be awesome!

ryanplusplus:
> Looks good!
> 
> If you could apply that spell fix to the exemplar as well so would that be awesome!

Good catch! Fixed.

meatball133 (approved):

PR #422, opened 2023-06-10
[New practice exercise]: Queen attack

ryanplusplus:


exercises/practice/queen-attack/.meta/src/example.cr
```diff
@@ -0,0 +1,24 @@
+class Queen
+  protected getter row, column
+
+  def initialize(@row : Int8, @column : Int8)
```
meatball133, line 4: I really like the use of Int8 here.

meatball133 (approved):
Looks good!

PR #426, opened 2023-06-21
[New exercise]: Saddle points

meatball133:


exercises/practice/saddle-points/.meta/src/example.cr
```diff
@@ -0,0 +1,13 @@
+module SaddlePoints
+  def self.saddle_points(matrix : Array(Array(Int32))) : Array(NamedTuple(row: Int32, column: Int32))
+    result = Array(NamedTuple(row: Int32, column: Int32)).new
```
ryanplusplus, line 3: > NamedTuple

:100:

exercises/practice/saddle-points/.meta/src/example.cr
```diff
@@ -0,0 +1,13 @@
...
+    matrix.each_with_index do |row, i|
+      row.each_with_index do |value, j|
+        if value == row.max && value == matrix.transpose[j].min
```
ryanplusplus, line 6: Nice use of `transpose`!

exercises/practice/saddle-points/spec/saddle_points_spec.cr
```diff
@@ -0,0 +1,58 @@
+require "spec"
+require "../src/*"
+
+describe "SaddlePoints" do
+  it "Can identify single saddle point" do
+    input = [[9, 8, 7], [5, 3, 2], [6, 6, 7]]
+    expected = [{row: 2, column: 1}]
+    SaddlePoints.saddle_points(input).sort_by { |hash| [hash[:row], hash[:column]] }.should eq(expected.sort_by { |hash| [hash[:row], hash[:column]] })
```
ryanplusplus, line 8: This is a little scary looking. Could we we move this into a helper so that it's not in every test? Or maybe we could use sets to simplify things:

```suggestion
    SaddlePoints.saddle_points(input).to_set.should eq(expected.to_set)
```

This is nicer to look at but doesn't check the count. We could go a bit further and do:

```suggestion
    actual = SaddlePoints.saddle_points(input)
    actual.size.should eq(expected.size)
    actual.to_set.should eq(expected.to_set)
```

meatball133: Good call!

PR #427, opened 2023-06-24
[New exercise]: Alphametics

ryanplusplus:


exercises/practice/alphametics/spec/alphametics_spec.cr
```diff
@@ -0,0 +1,68 @@
...
+
+  pending "puzzle with ten letters" do
+    puzzle = "AND + A + STRONG + OFFENSE + AS + A + GOOD == DEFENSE"
+    solution = {
+      'A' => 5, 'D' => 3, 'E' => 4, 'F' => 7, 'G' => 8, 'N' => 0, 'O' => 2, 'R' => 1, 'S' => 6, 'T' => 9,
+    }
+    Alphametics.solve(puzzle).should eq(solution)
+  end
+
+  pending "puzzle with ten letters and 199 addends" do
```
ryanplusplus, line 61: This test takes ~12 seconds on my machine which I think will time out in the online editor. I ported my solution from Ruby where it takes ~5-6 seconds (which passes the editor). I'm going to see if there are any optimization opportunities in my solution or consider removing this test from the test suite.

meatball133: I can start up the test runner in docker and it took 20 seconds to run with a codespace session with 4 cores. I think that test has to be shortened since at least at the moment there is not a possibility for parallel execution.

I have had similar student encounters where they use combinations, permutation etc and complain about the slow speed.
What makes Crystal slow here is permutation, Ruby can do a huge permutation in a few milliseconds, and Crystal can require seconds, I can dig into why this is happening but this will likely not be fixed for a while.

ryanplusplus: Thanks for the info. I suspected it was something to do with permutations, but I didn't do any profiling to prove it. I'll disable this test for now.

meatball133 (approved):
Nice!

PR #429, opened 2023-06-25
Return set from SaddlePoints.saddle_points

ryanplusplus:
When I was doing this exercise I realized we could simplify things by returning a set of results instead of an array.

@meatball133 - what do you think about this approach?


meatball133 (approved):
Yeah, I think this seems like a good idea!

PR #434, opened 2023-07-06
[Johannes Juice Maker]: Improve on TDD 

meatball133:


exercises/concept/johannes-juice-maker/spec/johannes_juice_maker_spec.cr
```diff
@@ -10,77 +10,117 @@ describe "JuiceMaker" do
...
-      juice_maker = JuiceMaker.new(0)
-      juice_maker.@fluid.should eq 0
-      juice_maker.@running.should be_false
+      {% if JuiceMaker.has_method? "initialize" %}
+        juice_maker = JuiceMaker.new(0)
+        juice_maker.@fluid.should eq 0
+        juice_maker.@running.should be_false
+      {% else %}
+        raise "Error: wrong number of arguments for 'JuiceMaker.new' (given 1, expected 0)"
+      {% end %}
```
ryanplusplus, line 19: Forgive my ignorance, but I'm not sure what I'm looking at here. Does `crystal_spec` evaluate this template at test time? Is there a resource you can point me to so that I can learn more? I tried to google this but I'm not sure what to search for.

meatball133: These are macros: https://crystal-lang.org/reference/1.8/syntax_and_semantics/macros/index.html
It more or less allows you to write code at compile time. This lets us get around the test-raising errors when not all methods are defined.
If the method exists then the test case looks like this:

```crystal
it "should create a new JuiceMaker" do
  juice_maker = JuiceMaker.new(0)
  juice_maker.@fluid.should eq 0
  juice_maker.@running.should be_false
end
```

If it doesn't exist:

```crystal
it "should create a new JuiceMaker" do
  raise "Error: wrong number of arguments for 'JuiceMaker.new' (given 1, expected 0)"
end
```

It is heavily used in the language files and will be used in the representer as well.

ryanplusplus: Thanks for the links!

ryanplusplus (approved):

PR #443, opened 2023-07-22
[New exercise]: Book store

ryanplusplus:


config.json
```diff
@@ -1034,6 +1034,14 @@
...
         "prerequisites": [],
         "difficulty": 6
+      },
+      {
+        "slug": "book-store",
+        "name": "Book Store",
+        "uuid": "9881c617-2388-4c30-884b-f79cbebcdb55",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 7
```
ryanplusplus, line 1044: This might look high given the length of the example solution, but this is a deceptively difficult problem. For reference, Ruby and Go both have this rated as an 8 and Rust has it rated as a 7.

meatball133: I think 7 is fine, since as you said it is a difficulty problem

meatball133 (approved):
Good job! ✨

PR #458, opened 2023-08-07
fix the spec to match the text

copiousfreetime:
The specs text says it should be 0, but tests for 4, and to match the rest of the examples, and probably the expected solution given that there are 4 sensors, I think the text is correct. At least that's how I read it from the README.

This test does determine the difference if you want `(4 - number_of_sensors) % 4` to be the solution, or `4 - number_of_sensors % 4`

Although according to the [examplar](https://github.com/exercism/crystal/blob/c081fc279ccf5e7d438f74da0dd9d8aa51fe3a1b/exercises/concept/wellingtons-weather-station/.meta/src/exemplar.cr) you want the latter solution to match.

In which case the text should be updated.

### DIFF
--- a/exercises/concept/wellingtons-weather-station/.meta/src/exemplar.cr
+++ b/exercises/concept/wellingtons-weather-station/.meta/src/exemplar.cr
@@ -12,6 +12,6 @@ class Temperature
   end
 
   def number_missing_sensors(number_of_sensors)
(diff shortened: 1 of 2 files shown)


meatball133:
I think this makes sense, but as you said the exemplar has to be updated.

ryanplusplus (approved):

PR #484, opened 2023-09-06
Fix Isogram.isogram? declaration

ls1955:
Because:
It is currently define as an instance method (Isogram#isogram?), which probably should be a class method judging from the spec.


ryanplusplus (approved):

ryanplusplus:
Thanks!

ls1955:
Glad I could help. :D

PR #487, opened 2023-09-17
[New Exercise]: Yacht

ryanplusplus:


meatball133 (commented):
I think the exercises look good, but I think an append is needed due to the usage of Enums. As the one found here: https://github.com/exercism/crystal/blob/main/exercises/practice/state-of-tic-tac-toe/.docs/instructions.append.md

ryanplusplus:
> I think the exercises look good, but I think an append is needed due to the usage of Enums. As the one found here: https://github.com/exercism/crystal/blob/main/exercises/practice/state-of-tic-tac-toe/.docs/instructions.append.md

Good call, I added a brief append

meatball133 (approved):

PR #493, opened 2023-09-25
[New Exercise]: Simple Cipher

kahgoh:
Looking to add the Simple Cipher exercise for [Object Oriented October](http://forum.exercism.org/t/select-featured-exercises-for-object-oriented-october/7112/18).

### DIFF
--- a/exercises/practice/simple-cipher/src/simple_cipher.cr
+++ b/exercises/practice/simple-cipher/src/simple_cipher.cr
@@ -1,13 +1,11 @@
 class SimpleCipher
+  # Write your code for the 'Simple Cipher' exercise in this file.
   def key
-    raise "Please implement the key method"
   end
 
   def encode(text : String)
-    raise "Please implement the encode method"
   end
 
   def decode(text : String)
-    raise "Please implement the decode method"
   end
 end
--- a/exercises/practice/simple-cipher/src/simple_cipher.cr
+++ b/exercises/practice/simple-cipher/src/simple_cipher.cr
@@ -1,0 +1,13 @@
+class SimpleCipher
+  def key
+    raise "Please implement the key method"
+  end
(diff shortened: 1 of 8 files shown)


meatball133:
To simplify the exercise creations, we generate from a template file, it often also makes the creation of the test file faster compared to writing it manually. And we can make changes with very little effort. Here is a doc on the Crystal test generator: https://github.com/exercism/crystal/blob/main/docs/generator.md. When writing the template, I recommend looking at how the other templates have been written and we do use ecr, which is the built-in template engine in Crystal. So there are no dependencies for running the script.

meatball133:
I could write you a template for this exericse if you like, but I would recommend to try it yourself.

kahgoh:
I'll have a go at creating the template ;)

exercises/practice/simple-cipher/src/simple_cipher.cr
```diff
@@ -0,0 +1,13 @@
+class SimpleCipher
+  def key
+    raise "Please implement the key method"
```
meatball133, line 3: Just one tiny nit, I think writing the stub file like this is nothing wrong with, and is the style for the concept exercises. However, the practice exercises don't use this style but only have a comment and don't use raise. like here: https://github.com/exercism/crystal/blob/main/exercises/practice/atbash-cipher/src/atbash_cipher.cr.

So for consistency, I would like that style to be used:
```
class SimpleCipher
   # Write your code for the 'Simple Cipher' exercise in this file.
  def key
  end

  def encode(text : String)
  end

  def decode(text : String)
  end
end
```

meatball133 (commented):
Nice job with the template, looking great, just one tiny nit

meatball133 (approved):

PR #500, opened 2023-09-29
Use standard lib methods

### LINKED ISSUE #405: [Templates]: Use standard-libary methods
We have implemented some custom methods that are already in the core language, to make the templates easier to maintain, there is a need to start using the standard methods.

This involves going through all of the practice exercise templates and replacing `to_capitalized` with the built-in method `camelcase`.
And `to_snake` with the built in method `underscore`.

ls1955:
This commit:
* Refactor GeneratorHelp#to_capitalized
* Remove GeneratorHelp#to_snake
* Replace GeneratorHelp#to_snake with String#underscore across test_template.ecr

Associated Issue:
Closes #405

### DIFF
--- a/bin/generator_help.cr
+++ b/bin/generator_help.cr
@@ -101,16 +101,8 @@ class GeneratorHelp
     end
   end
 
-  def to_snake(input)
-    result = ""
-    input.each_char do |x|
-      result += x.upcase == x ? "_#{x.downcase}" : x
-    end
-    result
-  end
-
   def to_capitalized(input)
-    input.tr("-","_").camelcase
+    input.tr("-", "_").camelcase
   end
 
   def status
--- a/bin/generator_help.cr
+++ b/bin/generator_help.cr
@@ -101,6 +101,14 @@ class GeneratorHelp
     end
   end
 
+  def to_snake(input)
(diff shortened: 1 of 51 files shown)


ls1955:
Hmm... Can't figure out the error within space-age exercise.

ls1955:
After taking a look around, It might be because we are dealing with int and float in space-age exercise?

```Crystal
num_one = 1000000000.to_s
num_two = 31.69.to_s

to_snake(num_one) # => "_1_0_0_0_0_0_0_0_0_0"
num_one.underscore # => "1000000000"
to_snake(num_two) # => "_3_1_._6_9"
num_two.underscore # => "31.69"
```

bin/generator_help.cr
ryanplusplus, line 1: We can remove `to_snake` now, right?

ryanplusplus (changes requested):

bin/generator_help.cr
ryanplusplus: Never mind, I see it's still used in `space-age`. Let me see if I can find a way to use `String#underscore` in `space-age`.

ryanplusplus (approved):

PR #516, opened 2023-10-07
[New Exercise]: Complex Numbers

ryanplusplus:


exercises/practice/complex-numbers/.meta/test_template.ecr
```diff
@@ -0,0 +1,77 @@
...
+                <% if cases["cases"]? %>
+                        <% flattened_cases.concat(cases["cases"].as_a) %>
+                <% else -%>
+                        <% flattened_cases.concat([cases]) %>
+                <% end %>
+            <% end %>
+        <% else -%>
+            <% flattened_cases.concat([cases]) %>
+        <% end %>
+    <% end %>
```
ryanplusplus, line 43: The structure of the canonical data is not friendly for automation. I added this to flatten the structure so I wouldn't need to duplicate the logic for creating each test case (which is already very complex) once for every level of nesting.

exercises/practice/complex-numbers/.meta/test_template.ecr
```diff
@@ -0,0 +1,77 @@
...
+<%
+complex_args = ->(json : JSON::Any) {
+  json
+    .as_a
+    .map(&.to_s)
+    .map(&.gsub("e", "Math::E"))
+    .map(&.gsub("pi", "Math::PI"))
+    .map(&.gsub("ln", "Math.log"))
+    .join(",")
+}
```
ryanplusplus, line 25: I had to use a proc literal for this because I was getting errors trying to use `def`. I'm not sure if this is expected or if I was doing something wrong.

exercises/practice/complex-numbers/.docs/instructions.append.md
```diff
@@ -0,0 +1,4 @@
+In this exercise you will need to add methods to an existing struct (`Number`).
+You can see an example of this in the Crystal documentation for [getters and setters][getters-and-setters]
+
+[getters-and-setters]: https://crystal-lang.org/reference/1.9/syntax_and_semantics/methods_and_instance_variables.html#getters-and-setters
```
ryanplusplus, line 4: Since this isn't a section specifically about reopening/extending classes, I chose to link to a specific version (1.9) so that the link would be less likely to be incorrect in the future.

exercises/practice/complex-numbers/.meta/test_template.ecr
meatball133: I think it is fine, I have seen the problem spec repo for this exercise and it is structured in a way that makes it a bit complicated.

meatball133: If you like you could add a method here if you feel like the template becomes a bit complicated: https://github.com/exercism/crystal/blob/main/bin/generator_plugins.cr

meatball133 (approved):

exercises/practice/complex-numbers/.meta/test_template.ecr
ryanplusplus: I'll do that. I realize now that this uses ECR in a way that's unnecessary and it could instead just use Crystal.

PR #528, opened 2023-10-15
Handle numerical issues complex_numbers_spec.cr

f-fr:
The spec uses `eq` to test for equality of complex numbers. However, the real and imaginary parts are `Float64`, hence there might be numerical issues. Instead, the `be_close` expectation should be used.

Using `eq` would require the `Complex` class to be modified (e.g. by overloading `==` to be slightly imprecise). But this would somehow hard-code the precision of `Complex`, which would not be a good idea if `Complex` was a "production" implementation. Although it is only an exercise, it should not encourage "bad style" like this.

### DIFF
--- a/exercises/practice/complex-numbers/.meta/src/example.cr
+++ b/exercises/practice/complex-numbers/.meta/src/example.cr
@@ -59,10 +59,6 @@ class Complex
   def exp
     self.class.new(Math.exp(real), 0) * self.class.new(Math.cos(imaginary), Math.sin(imaginary))
   end
-
-  def ==(other : Complex)
-    (real - other.real).abs < 1e-9 && (imaginary - other.imaginary).abs < 1e-9
-  end
 end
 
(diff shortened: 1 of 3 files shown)


ryanplusplus (approved):

ryanplusplus:
The changes look good but they need to be done in the test template because the tests aren't generated instead of manually written.

ryanplusplus (changes requested):
Make these changes in `exercises/practice/complex-numbers/.meta/test_template.ecr`

f-fr:
I could also modify the example solution and remove the `==` operator (it is not needed anymore).

ryanplusplus:
> I could also modify the example solution and remove the `==` operator (it is not needed anymore).

Yes please

ryanplusplus (approved):

meatball133 (approved):

PR #544, opened 2023-10-26
Adds test case for nil in resistor color

johnsyweb:
#### Context

When completing https://exercism.org/tracks/crystal/exercises/resistor-color , the signature for `def self.color_code(color : String) : Int32 | Nil` shows that the method can return `nil` but there's no spec to describe _why_.

#### Change

- Add test for nil return case
- Improve output for `crystal spec -v`

#### Considerations

0f100515 is not required but I think it improves the output.

<details>
  <summary>Before: `crystal spec -v `</summary>

```                      
ResistorColor
  Black
  White
  Orange
  handles input that is not a recognised color
  Colors

Pending:
  ResistorColor White
  ResistorColor Orange
  ResistorColor handles input that is not a recognised color
  ResistorColor Colors

Failures:

  1) ResistorColor Black
     Failure/Error: ResistorColor.color_code("black").should eq(0)

       Expected: 0
            got: nil

     # spec/resistor_color_spec.cr:6

Finished in 331 microseconds
5 examples, 1 failures, 0 errors, 4 pending

Failed examples:

crystal spec spec/resistor_color_spec.cr:5 # ResistorColor Black
```

</details>

<details>
  <summary>After `crystal spec -v`</summary>

```                          
ResistorColor
  .color_code
    finds the value for black
    finds the value for white
    finds the value for orange
    handles input that is not a recognised color
  .colors
    returns an array of known colors

Pending:
  ResistorColor .color_code finds the value for white
  ResistorColor .color_code finds the value for orange
  ResistorColor .color_code handles input that is not a recognised color
  ResistorColor .colors returns an array of known colors

Failures:

  1) ResistorColor .color_code finds the value for black
     Failure/Error: ResistorColor.color_code("black").should eq(0)

       Expected: 0
            got: nil

     # spec/resistor_color_spec.cr:7

Finished in 299 microseconds
5 examples, 1 failures, 0 errors, 4 pending

Failed examples:

crystal spec spec/resistor_color_spec.cr:6 # ResistorColor .color_code finds the value for black
```

</details>

### DIFF
--- a/exercises/practice/resistor-color/spec/resistor_color_spec.cr
+++ b/exercises/practice/resistor-color/spec/resistor_color_spec.cr
@@ -1,24 +1,28 @@
 require "spec"
 require "../src/*"
 
-describe "ResistorColor" do
-  it "Black" do
-    ResistorColor.color_code("black").should eq(0)
+describe ResistorColor do
+  describe ".color_code" do
+    it "finds the value for black" do
+      ResistorColor.color_code("black").should eq(0)
+    end
+
+    pending "finds the value for white" do
+      ResistorColor.color_code("white").should eq(9)
+    end
+
+    pending "finds the value for orange" do
+      ResistorColor.color_code("orange").should eq(3)
+    end
+
+    pending "handles input that is not a recognised color" do
+      ResistorColor.color_code("Puce").should be_nil
+    end
   end
 
-  pending "White" do
-    ResistorColor.color_code("white").should eq(9)
-  end
-
-  pending "Orange" do
-    ResistorColor.color_code("orange").should eq(3)
-  end
-
-  pending "handles input that is not a recognised color" do
-    ResistorColor.color_code("Puce").should be_nil
-  end
-
-  pending "Colors" do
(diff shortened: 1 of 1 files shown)


meatball133:
Hi!

I have thought about this addition, but first in for all, all changes need to be included here: https://github.com/exercism/crystal/blob/main/exercises/practice/resistor-color/.meta/test_template.ecr

To update the template's test file, just run: `bin/generator.sh resistor-color`.

But back to the addition, I am thinking about it and in more or less any language the type you will receive will always either be `nil` or `null` or int. The difference is that here is it explicitly written. I am arguing a bit with myself if it is more in place of making a pr to problem spec including this instead of a track-specific test.

But I suppose for now it could be a track-specific, to add the extra test you will have to in the meta directory add an `additional_tests.json` file with the extra test case wrapped in a json file like this:

```json
{"cases": {<test_case>}}
```

and then add the test case in the toml file.

johnsyweb:
Thanks for the review @meatball133 . Consistency with the other tracks is _probably_ more important and the expected behaviour for an invalid colour could vary between `nil`, `null` and some kind of exception. As such I'll close this one off.

PR #545, opened 2023-10-28
Add Pop Count

kahgoh:
This adds the [Pop Count exercise](https://forum.exercism.org/t/new-exercise-suggestion-pop-count/7632)

### DIFF
--- a/config.json
+++ b/config.json
@@ -657,7 +657,7 @@
         "slug": "pop-count",
         "name": "Pop Count",
         "uuid": "bdf29951-ece0-4212-ae1f-c4a09359aa45",
-        "practices": ["numbers", "while-loops"],
+        "practices": ["bit-manipulation"],
         "prerequisites": [],
         "difficulty": 3
       },
--- a/config.json
+++ b/config.json
@@ -654,6 +654,14 @@
         "difficulty": 3
       },
       {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "bdf29951-ece0-4212-ae1f-c4a09359aa45",
+        "practices": ["numbers", "while-loops"],
+        "prerequisites": [],
+        "difficulty": 3
+      },
+      {
         "slug": "proverb",
         "name": "Proverb",
         "uuid": "3cef1164-8eab-4e2d-a455-2e0630919bd1",
--- a/exercises/practice/pop-count/.meta/src/example.cr
+++ b/exercises/practice/pop-count/.meta/src/example.cr
@@ -1,0 +1,10 @@
+class PopCount
+  def self.egg_count(number : Int32) : Int32
+    count : Int32 = 0
+    while number > 0
+      count += (number & 1)
+      number = number >> 1
+    end
+    count
+  end
+end
--- a/exercises/practice/pop-count/src/pop_count.cr
+++ b/exercises/practice/pop-count/src/pop_count.cr
@@ -1,0 +1,5 @@
+class PopCount
+  def self.egg_count(number : Int32) : Int32
+    # Write your code for the 'Pop Count' exercise in this file.
+  end
+end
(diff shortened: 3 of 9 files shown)


kahgoh:
While working on this PR, I wonder if it was worth adding a concept for the bit manipulation operators? I'd be happy to have a go at adding the concept if we think it is.

config.json
```diff
@@ -653,6 +653,14 @@
         "prerequisites": [],
         "difficulty": 3
       },
+      {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "bdf29951-ece0-4212-ae1f-c4a09359aa45",
+        "practices": ["numbers", "while-loops"],
```
meatball133, line 660: I think just putting "bit manipulation" as practice is better since we already have quite a few exercises practicing numbers and while loops.

meatball133 (commented):
Looks good just one tiny nit

meatball133:
>While working on this PR, I wonder if it was worth adding a concept for the bit manipulation operators? I'd be happy to have a go at adding the concept if we think it is.

I think a concept about bit manipulation is totally valid. So the operators: `<<`, `>>`, `&`, `^`, `|`, and if you find something more feel free to add.

I have thought about adding a binary, octal and hexadecimal, so I think this concept would link pretty nice to that one. I was thinking of putting the binary, octal, and hexadecimal concept after String in a new branch (not connected to concepts after strings). If you want I could iterate a binary, octal and hexadecimal concept and then it maybe would be easier to work on the bit manipulation concept?

config.json
kahgoh: I wasn't sure about this one as the "bit manipulation" concept hasn't been written yet. Originally, I was thinking of doing the concept and updating this `practice` array together in a later PR. I'm happy to still change this now or add the concept to this PR if either is still preferred.

kahgoh:
> I have thought about adding a binary, octal and hexadecimal, so I think this concept would link pretty nice to that one. I was thinking of putting the binary, octal, and hexadecimal concept after String in a new branch (not connected to concepts after strings). If you want I could iterate a binary, octal and hexadecimal concept and then it maybe would be easier to work on the bit manipulation concept?

That sounds useful!

config.json
meatball133: I think doing it now is better and then we can add a concept later.

meatball133 (approved):

exercises/practice/pop-count/src/pop_count.cr
```diff
@@ -0,0 +1,5 @@
+class PopCount
```
ryanplusplus, line 1:
```suggestion
module PopCount
```

exercises/practice/pop-count/.meta/src/example.cr
```diff
@@ -0,0 +1,10 @@
+class PopCount
```
ryanplusplus, line 1:
```suggestion
module PopCount
```

exercises/practice/pop-count/.meta/src/example.cr
```diff
@@ -0,0 +1,10 @@
...
+  def self.egg_count(number : Int32) : Int32
+    count : Int32 = 0
```
ryanplusplus, line 3:
```suggestion
    count = 0_i32
```

I'm fine either way, but this is an alternative that you may prefer

exercises/practice/pop-count/.meta/src/example.cr
```diff
@@ -0,0 +1,10 @@
...
+    while number > 0
+      count += (number & 1)
+      number = number >> 1
```
ryanplusplus, line 6:
```suggestion
      number >>= 1
```

ryanplusplus (changes requested):

exercises/practice/pop-count/.meta/src/example.cr
meatball133: Crystal will assign by default int32 so I think it is also fine to just write it as `count = 0`.

meatball133:
Well it is already merged, but if you like you can open a new pr with the fixes.

PR #553, opened 2023-11-04
[New Exercise]: Word Search

ryanplusplus:


exercises/practice/word-search/.meta/src/example.cr
ryanplusplus, line 1: I'm not a big fan of how this turned out. It was pretty straightforward for horizontal and vertical but it got way out of hand for the diagonals. I'm going to try the more direct approach where I check each starting point and iterate in all directions.

meatball133 (approved):

PR #573, opened 2024-01-05
[Reverse-String]: Add approaches

meatball133:


exercises/practice/reverse-string/.approaches/introduction.md
```diff
@@ -0,0 +1,66 @@
...
+This is the most exemplary and straightforward solution.
+It uses the built-in [`reverse` method][reverse-method] to reverse the string.
+
+```crystal
+module ReverseString
+  def self.reverse(input : String) : String
+    input.reverse
+  end
+end
+```
```
ErikSchierboom, line 22: No separate approach page for this? Even if it is very simple, I would still recommend creating one.

exercises/practice/reverse-string/.approaches/config.json
```diff
@@ -0,0 +1,34 @@
...
+      "blurb": "Uses String::Builder to build the string.",
+      "authors": [
+        "meatball133"
+      ]
+    },
+    {
+      "uuid": "f6cf9f18-47a6-408a-933c-40dd0536eef2",
+      "slug": "reverse",
+      "title": "Uses Reverse",
+      "blurb": "Uses the buult in reverse method.",
```
ErikSchierboom, line 30:
```suggestion
      "blurb": "Uses the built-in reverse method.",
```

exercises/practice/reverse-string/.approaches/reverse/content.md
```diff
@@ -0,0 +1,16 @@
...
+```crystal
+module ReverseString
+  def self.reverse(input : String) : String
+    input.reverse
+  end
+end
+```
+
+This is the simplest and most exemplary solution, using the built-in [`String#reverse`][reverse] method.
+It uses a very optimized implementation, using a buffer and working directly on the string's bytes.
```
ErikSchierboom, line 12:
```suggestion
Internally, it uses a very optimized implementation, using a buffer and working directly on the string's bytes.
```

ErikSchierboom (approved):
Nice! Just some tiny nits.

PR #588, opened 2024-01-19
[New Tool]: Doc Generator

meatball133:


.github/workflows/doc-generator.yml
```diff
@@ -0,0 +1,35 @@
...
+    types:
+      - opened
+      - synchronize
+
+jobs:
+  spec_doc_generator:
+    name: Check Generator Templates
+    runs-on: ubuntu-22.04
+    container:
+      image: crystallang/crystal
```
ErikSchierboom, line 14: Is it possible to pin this to a specific version?

meatball133: yes it is possible

.github/workflows/tests.yml
```diff
@@ -42,3 +51,16 @@ jobs:
         run: apt-get update && apt-get install -y jq
       - name: Run tests
         run: ./bin/test-exercises.sh
+  exercises-windows:
+    name: Check exercises Crystal on Windows
+    runs-on: windows-2022
+    steps:
+      - name: Install Crystal
+        uses: crystal-lang/install-crystal@v1
```
ErikSchierboom, line 59: Could you use a hash to pin the workflow?

ErikSchierboom: Could you use this same setup method instead of:

```
container:
      image: crystallang/crystal
```

.github/workflows/doc-generator.yml
ErikSchierboom: Please do

.github/workflows/tests.yml
meatball133: This piece of code is already live. But no windows doesnt support container.

ErikSchierboom: I meant the other way around, instead of using the Docker image, use `crystal-lang/install-crystal@v1`

meatball133: Ohh yeah that would work

.github/workflows/tests.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     runs-on: windows-2022
     steps:
       - name: Install Crystal
-        uses: crystal-lang/install-crystal@v1
+        uses: crystal-lang/install-crystal@e4c9d197a5d1fa8329b55dadf309d06050ef5d07
```
ErikSchierboom, line 59: Maybe also use this in .github/workflows/doc-generator.yml?

meatball133: Sure

ErikSchierboom (approved):

PR #591, opened 2024-01-23
[Raindrops]: Add approaches

meatball133:


exercises/practice/raindrops/.approaches/introduction.md
```diff
@@ -0,0 +1,51 @@
...
+    result += "Plang" if count.divisible_by?(5)
+    result += "Plong" if count.divisible_by?(7)
+    result.blank? ? count.to_s : result
+  end
+end
+```
+
+For more information, check the [conditionals approach][approach-conditionals].
+
+## Approach: Using rules
```
ErikSchierboom, line 29: I'm not entirely sure about "rules", but I also can't really come up with a better alternative. I think the C approach that looks like this called it data-driven or something like that.

ErikSchierboom (approved):

ErikSchierboom (approved):
Neat!

PR #596, opened 2024-01-28
Add List Ops exercise

kahgoh:


config.json
```diff
@@ -1043,6 +1389,16 @@
...
+      },
+      {
+        "slug": "list-ops",
+        "name": "List Ops",
+        "uuid": "b4a24540-f4d8-4520-a1a0-14144ab68dbe",
+        "practices": [
+          "procs-blocks"
+        ],
+        "prerequisites": [],
+        "difficulty": 5
```
ryanplusplus, line 1401: Based on the difficulty ratings for other exercises on this track, I think a 3 or 4 would be more appropriate, especially since the provided skeleton includes most of what's needed to implement generics.

exercises/practice/list-ops/spec/list_ops_spec.cr
```diff
@@ -0,0 +1,119 @@
+require "spec"
+require "../src/*"
+
+describe "ListOps" do
+  describe "append entries to a list and return the new list" do
+    it "empty lists" do
+      list_ops = ListOps(Int32).new
+      list_ops.append([] of Int32, [] of Int32).should eq([] of Int32)
+    end
+    pending "list to empty list" do
```
ryanplusplus, line 10: Can you update the template so that there are newlines between tests?

```suggestion
    end

    pending "list to empty list" do
```

exercises/practice/list-ops/spec/list_ops_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+    pending "empty list to list" do
+      list_ops = ListOps(Int32).new
+      list_ops.append([1, 2, 3, 4], [] of Int32).should eq([1, 2, 3, 4])
+    end
+    pending "non-empty lists" do
+      list_ops = ListOps(Int32).new
+      list_ops.append([1, 2], [2, 3, 4, 5]).should eq([1, 2, 2, 3, 4, 5])
+    end
+  end
+  describe "concatenate a list of lists" do
```
ryanplusplus, line 23: Can you update the template so that there are newlines between `describe`s?

```suggestion
  end

  describe "concatenate a list of lists" do
```

exercises/practice/list-ops/.meta/src/example.cr
```diff
@@ -0,0 +1,59 @@
+class ListOps(T)
```
ryanplusplus, line 1: I think this would be better as a `module` without instance methods. It's strange to initialize an object without state in order to be able to use these functions.

test-generator/src/generator_help.cr
```diff
@@ -91,6 +91,22 @@ class GeneratorHelp
     input.tr("-", "_").camelcase
   end
 
+  def function(input : JSON::Any, arg_types : Array(String))
```
ryanplusplus, line 94: Can you add some documentation for this function? It wasn't clear to me what this was for until I looked at how it was used in the template.

meatball133, line 94: This method should likely be placed in the generator_plugins file istead since it is an addition to the templates.

kahgoh: Sounds good. I'll also move the two functions I had at the top of the template there too.

ryanplusplus (approved):

meatball133 (approved):

PR #613, opened 2024-03-16
[New Exercise] Food Chain

ryanplusplus:
Closes https://github.com/exercism/crystal/issues/609

### DIFF
--- a/config.json
+++ b/config.json
@@ -18,9 +18,6 @@
   "test_runner": {
     "average_run_time": 6
   },
-  "approaches": {
-    "snippet_extension": "cr"
-  },
   "files": {
     "solution": [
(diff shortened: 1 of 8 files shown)


config.json
```diff
@@ -18,9 +18,6 @@
   "test_runner": {
     "average_run_time": 6
   },
-  "approaches": {
-    "snippet_extension": "cr"
-  },
```
ryanplusplus, line 23: `bin/configlet` did this when I ran `create`. Is this section not actually used?

meatball133: It is not.
It is filled as an option in the docs but isn't yet implemented

meatball133 (approved):

PR #614, opened 2024-03-19
[New Exercise] D&D Character (#608)

adamazing:
Implements the D&D Character exercise in the Crystal track. Think I've covered off all the requirements per the docs.

- [x] Created test template for test generator `.meta/test_template.ecr`
  - [x] Generated `spec/dnd_character_spec.cr`
- [x] Created/updated `.meta/config.json`
- [x] Stubbed solution file at `src/dnd_character.cr`
- [x] Added an example that passes the spec at `.meta/src/example.cr`
- [x] Edited `config.json` to add exercise with doc generator and edited to add practices

Keen for any feedback you have 👍 

Issue: #608

### DIFF
--- a/config.json
+++ b/config.json
@@ -1422,6 +1422,17 @@
         ],
         "prerequisites": [],
         "difficulty": 3
+      },
+      {
+        "slug": "dnd-character",
+        "name": "D&D Character",
+        "uuid": "3ad6da95-f3ff-401b-83e7-6b8b7b29e133",
+        "practices": [
+          "classes",
+          "numbers"
+        ],
+        "prerequisites": [],
+        "difficulty": 1
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -1432,7 +1432,7 @@
           "numbers"
         ],
         "prerequisites": [],
-        "difficulty": 1
+        "difficulty": 2
       }
     ]
   },
--- a/exercises/practice/dnd-character/.meta/src/example.cr
+++ b/exercises/practice/dnd-character/.meta/src/example.cr
@@ -1,0 +1,32 @@
+class DndCharacter
+  property name : String
+  property strength : Int32
+  property dexterity : Int32
+  property constitution : Int32
+  property intelligence : Int32
+  property wisdom : Int32
+  property charisma : Int32
+  property hitpoints : Int32
+
+  def initialize(name : String = "Blandalf the Beige")
+    @name = name
+    @strength = self.class.ability
+    @dexterity = self.class.ability
+    @constitution = self.class.ability
+    @intelligence = self.class.ability
+    @wisdom = self.class.ability
+    @charisma = self.class.ability
+    @hitpoints = 10 + self.class.modifier(@constitution)
+  end
+
+  def self.modifier(score : Int32)
+    (((score - 10) / 2).floor).to_i
+  end
+
+  def self.ability : Int32
+    rolls = Array.new(4) { rand(1..6) }
+    rolls.sort!
+    rolls.shift
+    rolls.sum
+  end
+end
--- a/exercises/practice/dnd-character/src/dnd_character.cr
+++ b/exercises/practice/dnd-character/src/dnd_character.cr
@@ -1,0 +1,6 @@
+class DndCharacter
+  # Write your code for the 'DnD Character' exercise in this file.
+  def initialize(name : String = "Blandalf the Beige")
+    @name = name
+  end
+end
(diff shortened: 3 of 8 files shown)


adamazing:
Other tracks have tests like: 
```javascript
    xtest('ability score less than 3 throws error', () => {
      expect(() => abilityModifier(2)).toThrow(
        new Error('Ability scores must be at least 3'),
      );
    });

    xtest('ability score greater than 18 throws error', () => {
      expect(() => abilityModifier(19)).toThrow(
        new Error('Ability scores can be at most 18'),
      );
    });
  });
```
[Src](https://github.com/exercism/javascript/blob/12c896b6df32a47f16d9bb76fac40f9eb49b050b/exercises/practice/dnd-character/dnd-character.spec.js#L69-L80)

Should I add something similar?

ryanplusplus:
> Other tracks have tests like:
> 
> ```js
>     xtest('ability score less than 3 throws error', () => {
>       expect(() => abilityModifier(2)).toThrow(
>         new Error('Ability scores must be at least 3'),
>       );
>     });
> 
>     xtest('ability score greater than 18 throws error', () => {
>       expect(() => abilityModifier(19)).toThrow(
>         new Error('Ability scores can be at most 18'),
>       );
>     });
>   });
> ```
> 
> [Src](https://github.com/exercism/javascript/blob/12c896b6df32a47f16d9bb76fac40f9eb49b050b/exercises/practice/dnd-character/dnd-character.spec.js#L69-L80)
> 
> Should I add something similar?

My preference is to stick to the upstream problem specification. If it's a valuable test we should add it there.

@meatball133 what are your thoughts?

config.json
```diff
@@ -1422,6 +1422,17 @@
...
+      {
+        "slug": "dnd-character",
+        "name": "D&D Character",
+        "uuid": "3ad6da95-f3ff-401b-83e7-6b8b7b29e133",
+        "practices": [
+          "classes",
+          "numbers"
+        ],
+        "prerequisites": [],
+        "difficulty": 1
```
ryanplusplus, line 1435: This is an easy exercise, but I think the difficulty is probably a 2 or 3 based on the ratings we've given other exercises on this track.

exercises/practice/dnd-character/src/dnd_character.cr
```diff
@@ -0,0 +1,6 @@
+class DndCharacter
+  # Write your code for the 'DnD Character' exercise in this file.
+  def initialize(name : String = "Blandalf the Beige")
+    @name = name
+  end
```
ryanplusplus, line 5: Looks like `@name` is unused and unneeded. My vote is to remove it to avoid confusion.

exercises/practice/dnd-character/.meta/src/example.cr
```diff
@@ -0,0 +1,32 @@
...
+  property strength : Int32
+  property dexterity : Int32
+  property constitution : Int32
+  property intelligence : Int32
+  property wisdom : Int32
+  property charisma : Int32
+  property hitpoints : Int32
+
+  def initialize(name : String = "Blandalf the Beige")
+    @name = name
```
ryanplusplus, line 12: Looks like `@name` is unused and unneeded. My vote is to remove it to avoid confusion.

ryanplusplus (changes requested):
Looks great! Just a couple of nitpicks.

exercises/practice/dnd-character/src/dnd_character.cr
adamazing: I took a cue from [this bit](https://github.com/exercism/problem-specifications/blob/a57b3547b750d0f37af6897685fdbb81f2a5cf0b/exercises/dnd-character/canonical-data.json#L11-L12) in the `canonical-data.json` file:
 ```json
    ...
    "It may be fun to provide a 'name' property for characters",
    "that may or may not be randomly generated but not tested.",
    ...
 ```
 
 but happy to remove it too! 🙂

ryanplusplus: Ah, I didn't notice that!

@meatball133 what are your thoughts?

config.json
adamazing: Bumped to 2. I looked at a few other exercises to get an idea of where it sat, but my sample size may have been smaller and my perception of difficulty different. E.g. I saw that Roman Numerals was ranked 1 and assessed this exercise seemed to me to be conceptually easier than that at first blush?

Anyway, the JS track also ranks it as 2, so that seems like a good call between 2 and 3?

ryanplusplus: Works for me!

exercises/practice/dnd-character/src/dnd_character.cr
ryanplusplus: Let's stick with the name

ryanplusplus (approved):

PR #665, opened 2024-07-14
Update nil and add tag system

meatball133:
In prep for: https://github.com/exercism/crystal-test-runner/pull/85

Also allows to only run test for a certain test by just running `crystal spec --tag task_id=x`

### DIFF
--- a/concepts/nil/about.md
+++ b/concepts/nil/about.md
@@ -1,6 +1,6 @@
 # Nil
 
-Crystal has a type which is called [`Nil`][nil], it can only have one value: `nil`.
+Crystal has a type called [`Nil`][nil], which can only have one value: `nil`.
 It is used to represent the absence of a value, and is similar to `null` or `None` in other languages.
 
 Nil values can be returned from various methods for example [`String#[]?`][string-index], which returns `nil` if the index is out of bounds.
@@ -10,14 +10,14 @@ Nil values can be returned from various methods for example [`String#[]?`][strin
 ```
 
 Crystal has what is known as **NULL REFERENCE CHECKS**, which means that all types are non-nilable by default.
-This means that you can not assign `nil` to a variable unless you explicitly declare it as a `Nil` type.
+This means you can not assign `nil` to a variable unless you explicitly declare it as a `Nil` type.
 In turn this means that the compiler will automatically check for null references.
 
 ## Falsey value
 
-To be able to handle `nil` values, are there certain approaches that can be taken.
+Certain approaches can be taken to handle `nil` values.
 `nil` is a [falsey][truthy-falsey] value as well are `false`.
-This means that in if statements and other places where a falsey or truthy value is expected it will be treated the same as `false`.
+This means that statements and other places where a falsey or truthy value is expected it will be treated the same as `false`.
 
 ```crystal
 if nil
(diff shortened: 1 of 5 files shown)


exercises/concept/castle-dinner/.docs/instructions.md
```diff
@@ -2,18 +2,18 @@
...
 ## 1. Check if the food is correct
 
-There is a high chance that the food is poisoned if it is not the food that her majesty ordered.
-If the food is not the food that her majesty ordered, then it should be tossed out.
+There is a high chance that the food is poisoned if it is not what her majesty ordered.
+If the food is not what her majesty ordered, it should be tossed out.
 
 Implement the `check_food?` method, which takes the argument `food` which holds the name of the food as a `String`.
-The method should return the food if the food is `"Mushroom pasties"`, otherwise it should return `nil`.
+The method should return the food if it is' Mushroom pasties`; otherwise, it should return `nil`.
```
ryanplusplus, line 16:
```suggestion
The method should return the food if it is `Mushroom pasties`; otherwise, it should return `nil`.
```

I'm not sure this is what was intended, but something is not right.

exercises/concept/castle-dinner/.docs/instructions.md
```diff
@@ -57,3 +57,4 @@ CastleDinner.replace_drink("Tea")
 ```
 
 [kalmar_union]: https://en.wikipedia.org/wiki/Kalmar_Union
+
```
ryanplusplus, line 60:
```suggestion
```

exercises/concept/castle-dinner/.docs/introduction.md
```diff
@@ -53,12 +53,12 @@ end
 ```
 
 This is a bit different when just using the value in an if statement, since in the last examples would both false and nil be treated as false.
-Here only `nil` is treated as falsy, since if it was false it wouldn't have been nil thereby it would have been truthy.
+Here only `nil` is treated as falsey, since if it was false it wouldn't have been nil thereby it would have been truthy.
```
ryanplusplus, line 56:
```suggestion
Here only `nil` is treated as falsy, since if it was false it wouldn't have been nil thereby it would have been truthy.
```

concepts/nil/about.md
```diff
@@ -10,14 +10,14 @@ Nil values can be returned from various methods for example [`String#[]?`][strin
 ```
 
 Crystal has what is known as **NULL REFERENCE CHECKS**, which means that all types are non-nilable by default.
-This means that you can not assign `nil` to a variable unless you explicitly declare it as a `Nil` type.
+This means you can not assign `nil` to a variable unless you explicitly declare it as a `Nil` type.
 In turn this means that the compiler will automatically check for null references.
 
 ## Falsey value
```
ryanplusplus, line 16:
```suggestion
## Falsy value
```

concepts/nil/about.md
```diff
@@ -10,14 +10,14 @@ Nil values can be returned from various methods for example [`String#[]?`][strin
...
 
-To be able to handle `nil` values, are there certain approaches that can be taken.
+Certain approaches can be taken to handle `nil` values.
 `nil` is a [falsey][truthy-falsey] value as well are `false`.
```
ryanplusplus, line 19:
```suggestion
`nil` is a [falsy][truthy-falsy] value as well are `false`.
```

concepts/nil/about.md
```diff
@@ -53,12 +53,12 @@ end
 ```
 
 This is a bit different when just using the value in an if statement, since in the last examples would both false and nil be treated as false.
-Here only `nil` is treated as falsy, since if it was false it wouldn't have been nil thereby it would have been truthy.
+Here only `nil` is treated as falsey, since if it was false it wouldn't have been nil thereby it would have been truthy.
```
ryanplusplus, line 56:
```suggestion
Here only `nil` is treated as falsy, since if it was false it wouldn't have been nil thereby it would have been truthy.
```

I won't bother you by tagging all of these, but as far as I can tell "falsy" is far more common than "falsey"

meatball133: I think we have used falsey even though falsy is the most common because of this docs: https://crystal-lang.org/reference/1.13/syntax_and_semantics/truthy_and_falsey_values.html

Seems like the crystal developers calls it falsey so to keep it consistent with how they have wrote it so do I think falsey is better.

ryanplusplus: :+1:

ryanplusplus (approved):

PR #697, opened 2024-10-01
Add test to dominoes checking proper reversing

monsterkrampe:
I was mentoring a student and they submitted an implementation that passed all tests but was still not correct. In essence it allowed to use a domino in its reversed and unreversed form at the same time by only checking if one of the 2 positions matched any of the two other ones. While this would make the existing test fail, checking that `[1,2]` is not chainable, they also handled the case of only one domino separately, so this test actually passed.

In this PR I add another test that shall check if reversing is handled properly. The two dominoes `[1, 2], [3, 2]` are not chainable according to the exercise description.
(The solution submitted by the student fails on this test as expected.)


monsterkrampe:
Ah, I did not know that, thank you very much! :)

PR #706, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


ryanplusplus:
Would it make more sense to make the tests more flexible and not enforce an ordering?

IsaacG:
> Would it make more sense to make the tests more flexible and not enforce an ordering?

Probably! This PR is intended to communicate the current expectations. Feel free to close this PR if you'd rather leave it vague. Feel free to update the tests to be order-agnostic.

See the forum for the discussion: https://forum.exercism.org/t/14816/

ryanplusplus:
Sounds good, I will update the tests when I get a chance.

meatball133:
I am questionable about these changes. There are 10 other exercises with the same "problem", that the output order is specific (without any instructions saying it should be). And with these changes: https://github.com/exercism/problem-specifications/pull/2523/files I don't really see any real need for change or an appendum. The exercise could be changed to using sets instead (which would break all current solutions), but I am a bit questionable to other changes.

IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

IsaacG:
> I am questionable about these changes. There are 10 other exercises with the same "problem", that the output order is specific (without any instructions saying it should be). And with these changes: https://github.com/exercism/problem-specifications/pull/2523/files I don't really see any real need for change or an appendum. The exercise could be changed to using sets instead (which would break all current solutions), but I am a bit questionable to other changes.

This PR attempts to clarify what this track requires/expects. Having this clarification is certainly not required! Feel free to close out this PR if you'd prefer not to have this instruction append.

IsaacG:
@meatball133 Given the changes to the spec, would you like to:

1. merge this PR,
2. update the tests to change the behavior here, or
3. close this PR and not specify either way in the instructions?

IsaacG:
@meatball133  Ping

IsaacG:
@meatball133 Ping?

meatball133:
I still stand by my original stans, that "fixing" this on one exercise is a bit counterintuitive if every other exercise with this problem expects that you can understand the test file on the rest.

PR #707, opened 2025-02-05
Update to Crystal 1.15.1

meatball133:
Since I am planing to skip 1.16, so did I decide that I could just wait for a bug fix patch for 1.15 to make the experience a bit more stable.

### DIFF
--- a/.github/workflows/doc-generator.yml
+++ b/.github/workflows/doc-generator.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Install Crystal
         uses: crystal-lang/install-crystal@cdf26dcd488490c9939e9d4d62cab169c9e4f20d
         with:
-          crystal: "1.14.0"
+          crystal: "1.15.1"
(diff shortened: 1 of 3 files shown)


ryanplusplus (approved):

ryanplusplus:
> Since I am planing to skip 1.16

Just curious: why skip 1.16?

meatball133:
The language changes are mostly incremental because they update quarterly (meanwhile larger languages like Python and Ruby update yearly). And also when a new version drops, there is almost nothing that has to change. 1.15 had some breaking formatting changes but that doesn't change the language itself. So I don't think users notice new version even if they would use a newer version locally because the new code quite often as of recently works with old compilers.

 So I think doing it every 6 months makes more sense to make these kinds of updates.

Also I am currently working on the Swift track so I will not have any real meaning full content this spring for Crystal, so planning to instead make some content for the Summer and release it with 1.17.

PR #709, opened 2025-03-30
Added negative dividend modulo example

aduzsardi:


exercises/concept/wellingtons-weather-station/.docs/introduction.md
```diff
@@ -111,6 +114,13 @@ This is different from normal division.
 1 % 0
 # Error: Unhandled exception: Division by 0 (DivisionByZeroError)
 ```
+
+Note that unlike some other programming languages crystal is using floored division to calculate modulus (in JavaScript this would yield -1)
```
meatball133, line 118: Thing here is that, javascripts `%` isnt modulo, it is remainder, thereby another type of calculation.

aduzsardi: it is using the same operator so nomater what the type of calculation is it's still confusing for some people (as it was for me)

PR #739, opened 2026-01-01
Fix templates for new Crystal versions

meatball133:
Crystal seems to have changed how `inspect` works and I have opted for adding a new function which has similiar function to how inspect worked before.

### DIFF
--- a/test-generator/src/generator_help.cr
+++ b/test-generator/src/generator_help.cr
@@ -92,29 +92,6 @@ class GeneratorHelp
     input.tr("-", "_").camelcase
   end
 
-  def to_s_digged(input : JSON::Any) : String
-    case input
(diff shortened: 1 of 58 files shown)


test-generator/src/generator_help.cr
```diff
@@ -92,6 +92,29 @@ class GeneratorHelp
     input.tr("-", "_").camelcase
   end
 
+  def to_s_digged(input : JSON::Any) : String
```
ryanplusplus: What do you mean by "digged"?

ryanplusplus: How about `to_s_deep`? I think this will be clearer because it is similar to a shallow or deep comparison.

PR #746, opened 2026-01-03
Add grep exercise

meatball133:


bin/test-exercises.sh
```diff
@@ -7,6 +7,9 @@ echo $(crystal -v)
 test_run() {
     echo "Testing $1"
     cat "$1/.meta/src/$2.cr" > "${dic}/src/$2.cr"
+    if [ -d "$1/assets/." ]; then
+    cp -a "$1/assets/." "./assets/"
```
ryanplusplus:
```suggestion
        cp -a "$1/assets/." "./assets/"
```

exercises/practice/grep/src/grep.cr
ryanplusplus, line 1: This has the full implementation instead of a stub

ryanplusplus (approved):

PR #748, opened 2026-01-03
Add relative distance exercise

meatball133:


exercises/practice/relative-distance/.meta/src/example.cr
```diff
@@ -0,0 +1,63 @@
...
+
+class RelativeDistance
+  @graph : Hash(String, Set(String))
+
+  def initialize(family_tree : Hash(String, Array(String)))
+    @family_tree = family_tree
+    @graph = build_graph(@family_tree)
+  end
+
+  def degree_of_separation?(person_a : String, person_b : String) : Int32?
```
ryanplusplus:
```suggestion
  def degree_of_separation(person_a : String, person_b : String) : Int32?
```

Since this isn't a predicate, I don't think this should have a `?` in the method name.

ryanplusplus (approved):

